PR #6, opened 2018-11-27
Avoid usage of center_on_hmd()

Windfisch:
Instead, move the ARVROrigin appropriately. The rationale behind this
is to keep the ARVROrigin's local coordinate system intact, so that it
refers to the actual physical space.

Closes GodotVR/godot-vr-common#4


BastiaanOlij:
Looks good to me, I don't have time to try it out right now but I'll trust you tested this properly :)

PR #7, opened 2018-11-27
Scale the Player_figure according to the world_scale

Windfisch:
When altering the `world_scale`, the Player_figure size does not change accordingly. This patch allows to specify the player's height intuitively in gameworld units, and not in physical units.


BastiaanOlij:
Yeah makes sense to me, the only thing I can think could go wrong if someone already scales the player character in some other form. But I like the consistency of Godot by default making sure everything in the game world is in the game worlds scale while thing that deal with the real world are in the real world scale. So I'll merge this :)

PR #8, opened 2018-11-28
account for inverted PlaneMesh UVs

Windfisch:
godotengine/godot#23760 flipped the UVs in the PlaneMesh to be
consistent with QuadMesh.

FunctionTeleport uses PlaneMeshes, resulting in a wrong display
of the new view direction when teleporting, or the "marching
ants" indicating the teleport direction.

### DIFF
--- a/addons/vr-common/functions/Function_Teleport.tscn
+++ b/addons/vr-common/functions/Function_Teleport.tscn
@@ -45,7 +45,7 @@ void vertex() {
 void fragment() {
 	// and do our color
 	float offset =  (TIME * 2.0);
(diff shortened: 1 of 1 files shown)


Windfisch:
Note that this will cause the wrong orientation on old godot versions without godotengine/godot#23760. So merging / releasing this should be timed appropriately ;)

BastiaanOlij:
Might be nicer to just flip the image around?

BastiaanOlij:
I solved this differently and actually re-oriented the images. Made more sense to me than invert stuff.

Merged it with master, have a test :)

Windfisch:
That sounds sane :)

PR #11, opened 2019-03-10
Prevent hickup on first use of functions.

BastiaanOlij:
Adding a scene that prevents the hickup the user experiences the first time he/she uses certain functions.

Simply add the scene `misc/VR_Common_Shader_Cache.tscn` as a child node to your ARVRCamera node and this will fix this issue.


BastiaanOlij:
This also applies a bunch of cleanup:)

PR #14, opened 2019-08-24
Added Strafe and Fly movement

silverslade:
Modified the Direct Movement script
Added enum MOVEMENT_TYPE { MOVE_AND_ROTATE, MOVE_AND_STRAFE }
Added strafe action
Added fly action


BastiaanOlij:
ok, I'm going to merge this in because it works nicely. There is room for improvement but I think that could use a separate PR with some more though put into it.

I actually tried this out with the index instead of the rift and the button controls for flying don't feel very natural on that controller.

That's a matter of thinking about different modes. In similar fashion instead of flying a jump feature would be nice.

silverslade:
Great!

PR #21, opened 2020-03-09
Use is_instance_valid to check if targets are still valid

jtank4:
I experienced an error with Function_pointer.gd: After changing scenes based on a button on a 2din3d viewport UI being pressed (this viewport being part of the scene that was changed), the target variable was an already freed instance. Just doing a "if target" test is apparently insufficient to determine if the node was still there but replacing it with an is_instance_valid test fixed the issue.


BastiaanOlij:
Thanks! seems like a great idea to me!

PR #27, opened 2020-04-26
Added functionality to highlight and snap

BastiaanOlij:
Added functionality to highlight an object in range of being picked up and added a snap to location when picking up.

There are two new properties on the `Object Pickable` subscene:
image

`Center Pickup On` only works if `Reset Transform On Pickup` is true. If so the object will positioned so the controller is centered on the indicated node when the object is picked up.

image

image

`Highlight Mesh Instance` provides a mechanism where you can indicate the MeshInstance in your subscene that  represents the object being picked up. When the object is recognised as being the object closest to the hand the material entries for this MeshInstance will be temporarily replaced by a material that shows the object in highlight. 

image

### DIFF
--- a/addons/godot-xr-tools/VERSIONS.md
+++ b/addons/godot-xr-tools/VERSIONS.md
@@ -1,3 +1,8 @@
+2.1 (in progress)
+=================
+- added option to highlight object that can be picked up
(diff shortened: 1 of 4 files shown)


BastiaanOlij:
https://youtu.be/wdMLvWkLBXg

PR #32, opened 2020-07-05
In Godot 3.2 a new property hint was added that triggers the layer UI

BastiaanOlij:
Change the layer settings to show the proper UI:

image


BastiaanOlij:
Did a bit more tweaking to make the teleport function more stable, also added a feature that you can specify a max slope on which you can teleport

PR #36, opened 2020-10-30
Adding hand controllers that react on trigger and grip inputs

BastiaanOlij:
This PR adds scenes to our tools where we've implemented hand meshes that react to controller input.

Just add the scenes to an ARVRController node.


BastiaanOlij:
This now has both left and right hand thanks to @SYBIOTE , probably time to merge it in unless someone has suggestions for further improvements?

PR #41, opened 2021-04-10
Normalize after multiplication of movement

Noobidoo:
This is to stop acceleration if leaning back from original position.
This looks to fix that issue, and it's also how strafe movement is implemented


BastiaanOlij:
Not sure about this one, this turns the analogue joystick into an on/off control, not everyone will want that. In the current implementation you have fine control on how fast you want to move.

Noobidoo:
Hmm didn't notice that.
Have to do some further testing.

Henodude:
I think this PR should be closed, as it has had no activity for over a year and a half.

BastiaanOlij:
Decision in XR meeting, closing this as stale. Please raise a new issue if this is still relevant explaining more clearly what the problem is being encountered and why it should be fixed this way.

PR #42, opened 2021-06-10
Porting to Godot 4

BastiaanOlij:
Started on porting this to Godot 4, did most of the code changes and renames needed however a few things in Godot 4 are still broken so I'm unable to test this.


BastiaanOlij:
Direct movement and teleport now pretty much work, still working on the other features.

BastiaanOlij:
Updated this to include all the new changes that were in the Godot 3 version. Still a few things left to fix, it doesn't like extending `PhysicsBody3D` and it doesn't like `MovementProvider` and `PlayerBody` being dependent on each other, so still figuring out workarounds for that...

BastiaanOlij:
Ok, I worked around some of the issues that Godot 4 had, I haven't tested all things yet but I think nearly everything is converted.

Vignette is broken because `POSITION` seems broken in the shader.

BastiaanOlij:
I've merged this into an official `4.0-dev` branch now that Godot 4 XR is working.

PR #54, opened 2021-12-20
Add scrolling to Viewport_2D_in_3D

jakobbouchard:
I needed a way to scroll a long container with buttons, and thought it might be a nice addition to the tools!

It scrolls when the joystick is up or down (if more than 0.25 in value). If it stops scrolling, it sends a signal, to "unpress" the BUTTON_SCROLL_DOWN and BUTTON_SCROLL_UP. I found if I didn't put that it messed up the rest of the controls afterwards. If there's a better way to implement it, I'd be happy to use that instead!

### DIFF
--- a/addons/godot-xr-tools/functions/Function_pointer.gd
+++ b/addons/godot-xr-tools/functions/Function_pointer.gd
@@ -35,6 +35,7 @@ export var collide_with_areas = false setget set_collide_with_areas
 var target = null
 var last_target = null
 var last_collided_at = Vector3(0, 0, 0)
+var scrolling = null
 
 var ws = 1.0
 
@@ -142,6 +143,7 @@ func _process(delta):
 	
 	if enabled and $RayCast.is_colliding():
 		var new_at = $RayCast.get_collision_point()
+		var top_bottom = get_parent().get_joystick_axis(1)
 		
 		if is_instance_valid(target):
 			# if target is set our mouse must be down, we keep "focus" on our target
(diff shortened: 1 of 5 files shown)


jakobbouchard:
Also, it scrolls a bit fast, and I'm wondering if there's a way to use the axis' strength to decide how fast we scroll maybe?

BastiaanOlij:
Hmm, the principle solution is good but I wonder if it can be made more elegant. The problem here is that we're solving behavior on the pointer that belongs in the viewport. But there may not be another way.
Let me think about it for a bit to see if I can come up with a suggestion that might be slightly cleaner, if I can't I'll probably merge it as is :)

addons/godot-xr-tools/functions/Function_pointer.gd
```diff
@@ -142,6 +143,7 @@ func _process(delta):
 	
 	if enabled and $RayCast.is_colliding():
 		var new_at = $RayCast.get_collision_point()
+		var top_bottom = get_parent().get_joystick_axis(1)
```
BastiaanOlij, line 147: this should be set with an export property, not hardcoded to 1

jakobbouchard: oh, right my solution also won't support horizontal scrolling if we don't do that. I had it hardcoded since I knew I wanted it to be up-down on the joystick. left-right was for rotating the camera

PR #56, opened 2022-01-15
Use distance_squared method in function_pickup comparisons

jasonmorgado:
I noticed that this code for calculating the distance for nearby objects was equivalent to the `distance_to()` method, but that could also be replaced with `distance_squared_to()` since we are only using the distances for comparisons. 

As the [documentation](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-distance-squared-to) for the `distance_squared_to()` method describes, it's faster than the regular distance method and is preferred for situations like this.


BastiaanOlij:
Yup, saves a sqrt, not that important anymore these days but all little things help :) thanks!

Issue #58, opened 2022-01-19
Index fingers not working on LeftHand and RightHand


Malcolmnixon:
When using the LeftHand and RightHand assets, the grip button moves the middle, ring, and pinkly fingers; but the trigger does not move the index finger.

This is caused by a naming error in the filters of the blend trees in LeftHandBlendTree.tres and RightHandBlend.tres. The AnimationNodeBlend2 filters has an extra period in the index skeleton name which causes the blend-tree to not include any animation from the index finger branch in the final animation.

In both files the filters should be changed to remove the invalid period in the name as follows:
```
    - filters = [ "Armature_Left/Skeleton:index._distal", "Armature_Left/Skeleton:index._middle", "Armature_Left/Skeleton:index._proximal" ]
    + filters = [ "Armature_Left/Skeleton:index_distal", "Armature_Left/Skeleton:index_middle", "Armature_Left/Skeleton:index_proximal" ]
```

PR #59, opened 2022-01-20
Fixed filters of AnimationNodeBlend2 to match the bone names

Malcolmnixon:
Fixed filters of AnimationNodeBlend2 to match the bone names in the GLB files. This change allows the AnimationNodeBlend2 to correctly merge the index-finger animation driven by the analog-trigger input with the grip animation driven by the analog-grip input.

This fixes issue #58


BastiaanOlij:
Hmm, they must have changed with the importer changes in Godot 3.4. I actually want to replace this implementation with a better one that makes better use of the animation tree, need to find some time to implement it.
Anyway, good fix for now :)

PR #61, opened 2022-01-24
Modified Function_Direct_movement to use kinematic-body collision

### LINKED ISSUE #60: Direct Movement function freezes on slopes or steps
When using with the Direct Movement function, the player motion freezes when encountering steps, or slopes greater than around 35 degrees. The player then slides back down until control returns.

This appears to be caused by a RayCast "Tail" used for detecting whether the player is in contact with the ground:
image

Unfortunately this technique doesn't work with slopes or steps:
image image

Malcolmnixon:
Modified Function_Direct_movement to use kinematic-body collision and slope measurement testing rather than ray-cast tail to detect if the player is on the ground.

In addition a new drag coefficient has been added for when the player is in the air. This allows for more natural running off platforms.

This change resolves #60.


BastiaanOlij:
This looks like a nice improvement on the logic. I want to test it out but will likely merge it this week, thanks for the help Malcolm. Don't forget to add yourself to the CONTRIBUTORS.md file and add an entry in the CHANGES.MD file

Malcolmnixon:
I spent the weekend trying this out in a game-jam to test out the mechanics. The biggest take-away was the need for a system where the player mechanics can change in different areas - for example:
- The player steps onto ice and the ground-traction wants to drop to near-zero
- The player steps onto soft ground and the jump mechanic wants to be limited or disabled
- The player flies into a water-spout and the wind mechanic wants to increase its drag to pull the player along

I hacked something 'inelegant' for the jam, but a cleaner solution is to attach movement-modifier nodes to the ground objects, or to the wind areas. This would allow for setting overriding properties without having to 'scriptify' the object to contain the parameters.

Malcolmnixon:
This now includes the GroundPhysics node which can be attached to a ground object the player would walk along, and allows overriding of physical properties such as lowering traction to simulate ice, or limit/disable jumping.

I've uploaded a video to https://youtu.be/3uLxZ4BIR3U with a link to a demo project showing the capabilities.

BastiaanOlij:
I think its time we start moving to merging this before it becomes a monster :)

Couple of things, I noticed that `GroundPhysicsSettings` needs to become a tool script because it is used in other tool scripts and that is throwing errors in the editor and the property inspector stops working.

Then we need to make a decision on whether we're going to keep the exports for nodes. I'm for removing a bunch of them and just making everything declarative by placement of nodes. Might do this in a separate PR so people can give some feedback on it and there are a few places where we need to make the logic somewhat more intelligent, especially for selecting the camera node

Finally we need to squash all the commits before merging :)

Issue #64, opened 2022-02-04
Player 'drag' and 'traction' result in the player not moving the full speed


Malcolmnixon:
A PlayerBody has ground physics settings consisting of both 'drag' and 'traction' properties:
![image](https://user-images.githubusercontent.com/1863707/152454903-217f1f73-79c0-4264-ab85-7b0a41e6bf04.png)

Currently the rules for ground motion are:
1. **Always** slow the players velocity using the 'drag' parameter
2. If the player is trying to move, then Lerp the player velocity towards the target velocity using the 'traction' parameter

If we try to ascribe physical meaning to 'drag' and 'traction' we get:
* Drag is the passive control (to stationary) when the player isn't intentionally moving
* Traction is the active control when the player is trying to move

The descriptions above however are not perfectly correct because we apply 'drag' even if the player is attempting to move. The result of which is the player can't actually move at the 'Max Speed' because the 'drag' always pulls them slightly slower. For example the following screen-shot is the player set up with a 'Max Speed' of 4.0 running as fast as they can:
![image](https://user-images.githubusercontent.com/1863707/152455840-f30e281c-dc82-4057-b53d-ff6886c08f29.png)

One fix for this is to change the rules for ground motion as follows:
1. If the player is trying to move, then Lerp the player velocity towards the target velocity using the 'traction' parameter
2. If the player is **not** trying to move, then slow the velocity using the 'drag' parameter

Doing this simple change results in the player being able to run at the 'Max Speed' and still slowing if the joysticks are neutral:
![image](https://user-images.githubusercontent.com/1863707/152456111-43b6f3e6-d915-4363-b000-e9b8a7abc1fd.png)

BastiaanOlij:
Closing this as the solution is now merged :)

PR #65, opened 2022-02-04
Modified when ground-drag is applied

Malcolmnixon:
Modified ground-drag so it's only applied when the user isn't providing direct-control.

This fixes issue #64


BastiaanOlij:
Looks good to me, thanks!

PR #66, opened 2022-02-04
Added Virtual Keyboard

### LINKED ISSUE #46: Add virtual keyboard
Something that came up on the Godot XR discord channel, having a virtual keyboard implementation people can just use would be really handy

Malcolmnixon:
This pull request introduces a VirtualKeyboard2D scene, and a Virtual_keyboard scene which hosts it in a Viewport_2D_in_3D.

In addition, the generated key events weren't being passed in to controls hosted in a Viewport_2D_in_3D because of some esoteric parent/child viewport logic. The workaround is to have the Viewport_2D_in_3D process inputs, and manually pass them to the contained Viewport.

This change fixes #46

### DIFF
--- a/addons/godot-xr-tools/objects/Viewport_2D_in_3D.gd
+++ b/addons/godot-xr-tools/objects/Viewport_2D_in_3D.gd
@@ -81,9 +81,13 @@ func _ready():
 	set_scene(scene)
 	set_collision_layer(collision_layer)
 	set_transparent(transparent)
+	set_process_input(true)
 
 func _on_pointer_entered():
 	emit_signal("pointer_entered")
 
 func _on_pointer_exited():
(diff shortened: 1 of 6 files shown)


Malcolmnixon:
The changes that let keyboard events get routed through a Viewport_2D_in_3D to the contained controls and text-boxes also work on a Quest 2 with a paired bluetooth keyboard.

Testing was performed using a Logitech K830 keyboard (which the Quest 2 supports for keyboard passthrough); however the keyboard passthrough doesn't actually show - possibly due to the need for additional Android Manifest features.

Malcolmnixon:
Added alternate keys to get many of the additional characters.
image

BastiaanOlij:
Not sure what I did wrong originally but this is working like a charm for me now!

PR #67, opened 2022-02-04
Improve approach for hand blending

BastiaanOlij:
This PR changes the way we animate our hands based on trigger and grip inputs.

Instead of trying to pick frames in an animation, the hand now has different animation tracks that contain a single pose. Right now there are only two, an open hand and a closed hand. We could animate each pose a little to make the hands a little more alive but for now I've left that alone.

The model is just for the left hand and we mirror it for the right hand, this prevents us from having to maintain two models and allows us to reuse the animation tree. 

Note that the animation tree is greatly simplified:
image

We use our fist twice, ones to blend just the movement of the index finger based on our trigger input, and once to animate the pinky, ring finger and middle finger based on our grip input. It simply uses filters here.

We could easily add more gestures and add inputs that overrule things.


BastiaanOlij:
Cleaned up some old files and commented out the workaround for the trigger and grip values, if you use a master build of the openxr plugin it should work correctly.

BastiaanOlij:
@Malcolmnixon did you try this out ? I'm keen to merge it even though it does require a master build of the plugin.

Malcolmnixon:
@BastiaanOlij I tested https://github.com/GodotVR/godot_openxr/commit/7343d54ae340f45fd30176ea74cde1baf5e0eff4 and https://github.com/BastiaanOlij/godot-xr-tools/commit/4b19daf73475a9aa905433ab184b09adca5b05cd and the results look excellent on the Quest Native and Oculus Link via Oculus OpenXR Runtime.

BastiaanOlij:
Thanks Malcolm, I'll merge this and we'll pin something in issues for the time being about the wrong controller inputs.

Issue #68, opened 2022-02-05
Wind movement collision shape too low and doesn't always follow the player


Malcolmnixon:
The wind-sensor collision shape in the Function_Wind_movement Movement Provider is located at the players feet, and can get frozen in position if the player performs an exclusive movement operation such as climbing.

The following images show the wind-sensor when walking, then the wind-sensor when climbing:
![image](https://user-images.githubusercontent.com/1863707/152660939-60530663-0af8-4177-b216-5f597d4547bf.png)
![image](https://user-images.githubusercontent.com/1863707/152660966-569701b6-6443-4870-ae23-628b65361fc5.png)

The cause of the wind-sensor being at the players feet is because the Function_Wind_movement script places the wind-sensor at the PlayerBody/KinematicBody origin, which is not in the middle of the players body. The cause of the wind-sensor not following the player when an exclusive movement is occurring, is due to the exclusive movement skipping lower-priority movement providers, so the moving of the wind-sensor does not occur.

The simplest solution may just be to reparent the wind-sensor to the ARVRCamera.

BastiaanOlij:
Merged and thanks!

PR #69, opened 2022-02-05
Modified wind-sensor area to re-parent to the ARVRCamera

Malcolmnixon:
This change results in the wind-sensor area being attached to the players head, even in exclusive movement operations such as climbing:
image

This fixes issue #68


BastiaanOlij:
This looks fine to me, need to make sure the code is not executed in tool mode though

Malcolmnixon:
Added the editor check. Thanks for reviewing - it's taking me a little bit to build up the gdscript mental linting.

BastiaanOlij:
Thanks Malcolm!

Issue #70, opened 2022-02-05
Duplicated code to find player resource nodes


Malcolmnixon:
There are numerous almost-identical routines to find player resource nodes:
 - Find the ARVROrigin from a node under the player
 - Find the ARVRCamera from a node under the player
 - Find the PlayerBody from a node under the player

MovementProvider.gd has some of these functions, but they only work if the caller is a MovementProvider. Instead it would be better if these find routines were available to any node under the player - for example by making them static functions of a helper script.

BastiaanOlij:
It would indeed be nice to have a number of these functions in an autoload script, my main worry here is that this requires a setup action and I want as much as possible to remain automatic.

I wish GDScript had an include function :)

Malcolmnixon:
I added a pull request showing an example of what I'm thinking of. It lets a node located anywhere under the ARVROrigin do:
```
    var origin := ARVRHelpers.get_arvr_origin(self)
    var camera := ARVRHelpers.get_arvr_camera(self)
    var left_controller := ARVRHelpers.get_left_controller(self)
    var right_controller := ARVRHelpers.get_right_controller(self)
```

BastiaanOlij:
Yeah 100% agree with what it does, just thinking of how we can have the helper node there or auto loaded so the user of the plugin doesn't need to be instructed to do so before starting to use the plugin.

PR #71, opened 2022-02-06
Added ARVR_Helpers script.

Malcolmnixon:
This is a preliminary script related to issue #70 to help locate ARVR nodes.

I ended up not adding a function to look for the PlayerBody in here because it leads to circular dependencies. Instead my demo project adds the routine to the PlayerBody script:
```
    # Find the player body from our own node somewhere under the player
    var body := PlayerBody.get_player_body(self)
```

### DIFF
--- a/addons/godot-xr-tools/misc/ARVR_Helpers.gd
+++ b/addons/godot-xr-tools/misc/ARVR_Helpers.gd
@@ -1,0 +1,101 @@
+tool
+class_name ARVRHelpers
+
+
+##
+## ARVR Helper Rountines
+##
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Looks good, just need to find a nice way to make sure this is accessible. It should really be auto loaded, I wish there was a way to automate this.

BastiaanOlij:
Ok, I never knew this. It used to be so that you had to autoload scripts to use them, but the combo of using `class_name` with `static func` will automatically load the script if you call the static function from anywhere else in the code. :)

Issue #72, opened 2022-02-07
Picked up objects get the thrown velocity of the hand, not the object


Malcolmnixon:
The Function_Pickup script maintains an average velocity (linear and rotational) for itself, which is essentially the controller.

When an object is dropped, the controllers average linear and rotational velocity is applied to the object. This works only if the picked up object has "Reset Transform On Pickup" property enabled, which snaps the origin (center of mass) of the object to the Function_Pickup location.

If the "Reset Transform On Pickup" property is disabled to emulate precise-location grabbing, the origin (center of mass) of the object may be some distance from the controller, and the controllers linear and rotational velocity are not applicable.

Consider the scenario of throwing a long rod by holding one end, then flicking the wrist and letting go. As the hand was not necessarily moving, the rod will not move as expected.

The correct solution is to track the average velocity (linear and rotational) of the picked up object rather than the controller. In the example above of throwing a rod, the origin (center of mass) of the rod will be moving, and so when the rod is let go it will move with the expected velocity.

Malcolmnixon:
I believe the angular rotation averager should switch from Quaternion to Euler Angles; however it's going to take a bit of an explanation for why, including diving into how the physics engines interprets the Euler angular_velocity Vector3 value.

The Godot RigidBody class exposes angular_velocity as a Vector3 in Euler angles; however it should not be thought of as a rotation, but as a point angular velocity. The explanation for this comes from some code in the Godot 3 physics engine 'servers/physics/body_sw.cpp'. In there it takes the angular_velocity and does the following operations:
1. Measures the Vector3 length of the angular_velocity - this is treated as the a scalar angular velocity in radians per second
2. Normalize the Vector3 angular velocity - the result is always 1 radian in magnitude
3. Find the rotation axis described by the normalized angular velocity
4. Build a Basis using the rotation axis and the scalar angular velocity (scaled down from seconds to physics step-time)

Essentially these steps show the physics engine splits the incoming angular_velocity into two parts:
- The 'direction' of rotation (normalized angular_velocity) which is always 1 radian in magnitude
- The 'magnitude' of rotation which is encoded as the length of the angular_velocity vector

As such scaling the Euler 'angular_velocity' is a legitimate well-defined operation and will be extracted by the physics engine to control only the 'magnitude' of rotation.

[...]

PR #73, opened 2022-02-07
Thrown object velocity calculated from object instead of controller

Malcolmnixon:
This change resolves issue #72 by calculating the average velocity of the held object rather than the controller.


BastiaanOlij (approved):
Nice improvement over the old code :)

PR #88, opened 2022-03-03
Modified PlayerBody and movement functions to use new ARVRHelpers

Malcolmnixon:
Additional changes related to issue #70

PlayerBody.gd:
 - Changed NodePath export from default 'null' to 'NodePath()' as NodePath isn't friendly with nulls
 - Use ARVRHelpers to get origin and camera
 - Remove duplicate code found in ARVRHelpers
 - Added static PlayerBody::get_player_body() function

Function_Glide_movement.gd:
 - Changed NodePath export from default 'null' to 'NodePath()' as NodePath isn't friendly with nulls
 - Use ARVRHelpers to get controllers

Function_Wind_movement.gd:
 - Use ARVRHelpers to get camera
 - Remove duplicate code found in ARVRHelpers

MovementProvider.gd:
 - Use ARVRHelpers to get origin
 - Remove duplicate code found in ARVRHelpers

### DIFF
--- a/addons/godot-xr-tools/assets/PlayerBody.gd
+++ b/addons/godot-xr-tools/assets/PlayerBody.gd
@@ -38,17 +38,11 @@ export var push_rigid_bodies := true
 ## GroundPhysicsSettings to apply - can only be typed in Godot 4+
 export (Resource) var physics = null setget set_physics, get_physics
 
-## Path to the ARVROrigin node
-export (NodePath) var origin = null
-
-## Path to the ARVRCamera node
-export (NodePath) var camera = null
-
 ## ARVROrigin node
-var origin_node: ARVROrigin = null
+onready var origin_node := ARVRHelpers.get_arvr_origin(self)
 
 ## ARVRCamera node
-var camera_node: ARVRCamera = null
+onready var camera_node := ARVRHelpers.get_arvr_camera(self)
 
 ## Player KinematicBody node
 onready var kinematic_node: KinematicBody = $KinematicBody
@@ -91,44 +85,8 @@ class SortProviderByOrder:
 	static func sort_by_order(a, b) -> bool:
 		return true if a.order < b.order else false
 
-# Get our origin node, make sure we have consistent code here
-func _get_origin_node() -> ARVROrigin:
-	var node : ARVROrigin = get_node_or_null(origin) if origin else get_parent()
-	return node
-
-# Get our camera node
-func _get_camera_node() -> ARVRCamera:
-	# if we have set a node, try and use it
-	var node : ARVRCamera
-	
-	if camera:
-		node = get_node_or_null(camera)
-		if node:
-			return node
-
-	var o : ARVROrigin = _get_origin_node()
-	if !o:
-		return null
-
-	# else get by default name 
-	node = o.get_node_or_null("ARVRCamera")
-	if node:
-		return node
-
-	# else find the first camera child
-	for child in o.get_children():
-		if child is ARVRCamera:
-			return child
-
-	# no luck
-	return null
-
 # Called when the node enters the scene tree for the first time.
 func _ready():
-	# Get the origin and camera nodes
-	origin_node = _get_origin_node()
-	camera_node = _get_camera_node()
-
 	# Get the movement providers ordered by increasing order
 	_movement_providers = get_tree().get_nodes_in_group("movement_providers")
 	_movement_providers.sort_custom(SortProviderByOrder, "sort_by_order")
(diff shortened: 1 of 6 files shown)


addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -39,16 +39,16 @@ export var push_rigid_bodies := true
 export (Resource) var physics = null setget set_physics, get_physics
 
 ## Path to the ARVROrigin node
-export (NodePath) var origin = null
+export var origin := NodePath()
```
BastiaanOlij, line 42: I prefer `=` over `:=` personally. Also keeping in mind this will before `@export var origin : NodePath = NodePath()` in Godot 4 (though in Godot 4 it's not needed to add the `= NodePath()` when typed, the same might apply for Godot 3)
My personal preference though.

addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -39,16 +39,16 @@ export var push_rigid_bodies := true
...
 
 ## Path to the ARVRCamera node
-export (NodePath) var camera = null
+export var camera := NodePath()
 
 ## ARVROrigin node
-var origin_node: ARVROrigin = null
+onready var origin_node := ARVRHelpers.get_arvr_origin(self, origin)
```
BastiaanOlij, line 48: Just a warning for the future, I prefer onready var notation like this but Godot 4 has an interesting "bug" that may or may not get fixed. Having onready vars counts as having a `_ready` function even if you haven't added one to the script. If you have a `_ready` function you now have to call `super._ready()` or the parents ready function won't get called.

Doesn't impact Godot 3 but something to start keeping in mind for porting the code to our Godot 4 branch, which we should start doing so we don't deviate to far

addons/godot-xr-tools/assets/PlayerBody.gd
BastiaanOlij: There btw is also a new `@export_nodepath(type) var origin` notation in Godot 4 that allows you to limit what type of object you want to nodepath to point to but I think it still has some problems.

addons/godot-xr-tools/functions/Function_Glide_movement.gd
```diff
@@ -49,27 +49,21 @@ export var horizontal_slew_rate := 1.0
 export var vertical_slew_rate := 2.0
 
 ## Left ARVR Controller
-export (NodePath) var left_controller = null
+export var left_controller := NodePath()
```
BastiaanOlij, line 52: Should we keep these or should we just rely on get_left_controller and get_right_controller?

BastiaanOlij:
Looks good, most of my remarks were just related to the eventual port to Godot 4. Only thing we should probably discuss is whether we stick with the `:=` notation, I haven't seen many people use it with GDScript even though it is supported.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: For Godot3 it seems like the property-editor use both the typing system and export-hints; however the static-analysis typing system completely ignores export-hints. In other words:
```
export var some_int: int        # Export of explicitly typed integer, displays as int in the property-editor
export var some_int := 123      # Export of explicit (via assignment) typed integer, displays as int in the property-editor
export (int) var some_variant   # Export an untyped variant, but have it display as an int in the property-editor
```

So as far as I can tell ```export (NodePath) var origin``` would seem to be declaring an untyped variant without any code-editor hints, but shows correctly in the property-editor.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: Ugh, annoying but good to know. Languages are supposed to get easier over time and end up achieving more with less typing while at the same time defaulting to correct behavior. This sounds like a regression from Godot 3 that's going to mess people up.

When the language has the 'onready' keyword it's already an acknowledgement that _ready() isn't a normal function. Hopefully they at least add a static analysis warning/error if the _ready does not chain.

addons/godot-xr-tools/functions/Function_Glide_movement.gd
Malcolmnixon: The new JumpDetect just uses get_left_controller and get_right_controller, so I think these can go. I'll push an update with these removed.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: Maybe I'm getting a little too overly zealous about strong typing for Godot3 where all it can provide is some developer hints. For Godot4 I would really like to get the performance benefits of strong typing.

For Godot3 let's stick with ```export (NodePath) var origin```

Malcolmnixon: Removed numerous NodePath exports as no-longer needed, and the remaining ones use the preferred approach.

addons/godot-xr-tools/functions/Function_Glide_movement.gd
Malcolmnixon: Removed the unnecessary NodePath exports.

addons/godot-xr-tools/assets/PlayerBody.gd
BastiaanOlij: yeah i'm on the fence about the change too, but it does add in the ability to decide when the parent function is run, sometimes you want to be able to run logic before hand.

Issue #81, opened 2022-02-27
PlayerBody should expose collision settings


Malcolmnixon:
Currently to control the PlayerBody collision layers, "Editable Children" must be enabled, and the associated KinematicBody directly edited. Instead the collision settings should be accessible in the PlayerBody node.

PR #89, opened 2022-03-03
Added collision layer and mask exported properties

Malcolmnixon:
This pull request fixes issue #81 and contains:
 - Added collision layer and mask exported properties
 - Additionally added setget to radius to update the collision view in the editor

### DIFF
--- a/addons/godot-xr-tools/assets/PlayerBody.gd
+++ b/addons/godot-xr-tools/assets/PlayerBody.gd
@@ -21,10 +21,10 @@ extends Node
 ##
 
 ## PlayerBody enabled flag
-export var enabled := true setget set_enabled, get_enabled
+export var enabled := true setget set_enabled
 
 ## Player radius
-export var player_radius := 0.4
+export var player_radius := 0.4 setget set_player_radius
 
 ## Eyes forward offset from center of body in player_radius units
 export (float, 0.0, 1.0) var eye_forward_offset := 0.66
@@ -36,13 +36,19 @@ export var gravity := -9.8
 export var push_rigid_bodies := true
 
 ## GroundPhysicsSettings to apply - can only be typed in Godot 4+
-export (Resource) var physics = null setget set_physics, get_physics
+export (Resource) var physics = null setget set_physics
 
 ## Path to the ARVROrigin node
 export (NodePath) var origin = null
 
 ## Path to the ARVRCamera node
 export (NodePath) var camera = null
+
+# Set our collision layer
+export (int, LAYERS_3D_PHYSICS) var collision_layer = 1 << 19 setget set_collision_layer
+
+# Set our collision mask
+export (int, LAYERS_3D_PHYSICS) var collision_mask = 1023 setget set_collision_mask
 
 ## ARVROrigin node
 var origin_node: ARVROrigin = null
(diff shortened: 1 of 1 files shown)


addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -24,7 +24,7 @@ extends Node
 export var enabled := true setget set_enabled, get_enabled
 
 ## Player radius
-export var player_radius := 0.4
+export var player_radius := 0.4 setget set_player_radius, get_player_radius
```
BastiaanOlij, line 27: You can leave out the `get_player_radius` seeing it just gets the player radius, you can have property definitions with just a setter and Godot will use the default getter.

addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -44,6 +44,12 @@ export (NodePath) var origin = null
 ## Path to the ARVRCamera node
 export (NodePath) var camera = null
 
+# Set our collision layer
+export (int, LAYERS_3D_PHYSICS) var collision_layer = 1 << 19 setget set_collision_layer, get_collision_layer
```
BastiaanOlij, line 48: Seeing we've got this in a bunch of places now, I think we should make the default values a constant in our helper script. Something like:
```
const XR_TOOLS_PLAYER_LAYER = 1 << 19
const XR_TOOLS_DEFAULT_MASK = 1023
```
etc.

addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -147,6 +157,13 @@ func set_enabled(new_value):
 func get_enabled():
 	return enabled
 
+func set_player_radius(new_value: float) -> void:
+	player_radius = new_value
+	_collision_node.shape.radius = new_value
```
BastiaanOlij, line 162: ok be careful here, the first time this gets called when a script loads, `_collision_node` won't be initialised yet. You'll see in much of the code I use a pattern like this:
```
export var some_value setget set_some_value

func set_some_value(new_value):
    some_value = new_value
    if inside_tree():
      _update_some_value()

func _update_some_value():
    # do stuff here that needs doing

func _ready():
    # now that we're ready, we can propogate all our defaults
    _update_some_value()

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: So omitting one of the functions gives it 'default' behavior. I wish the effects of omitting setters and getters was clearer in the manual. I didn't know if omitting the setter meant default behavior or made the variable read-only, and if omitting the getter meant default behavior or making the variable write-only.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: The only place in code for these constants is now PlayerBody.gd (with this pull request). We do however have collision values encoded into scene nodes:
- PlayerBody/KinematicBody layer/mask (1<<19 and 1023 - now controlled by PlayerBody.gd)
- Function_Pickup layer/mask (1<<19 and 1023)
- Function_Pointer mask (1<<20 - 1)
- Function_Teleport layer/mask (1<<19 and 1023)
- Function_Wind_Movement mask (1<<20)

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: I've applied this pattern to all the appropriate setters.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: I've removed the unnecessary getters.

addons/godot-xr-tools/assets/PlayerBody.gd
BastiaanOlij: That's indeed the thing with declarative development through nodes, you can't really add in constants that way :)

BastiaanOlij:
@Malcolmnixon merging the other PR caused a merge conflict on this one, so if you could resolve that I can merged it :)

Issue #87, opened 2022-03-03
PlayerBody cannot be added to player after movement providers


Malcolmnixon:
The movement providers attempt to create a PlayerBody if one does not exist; however the PlayerBody does not appear in the Scene tree-view even though it exists. The result is that:
 1. The generated PlayerBody is not visible, so none of its settings can be modified
 2. Attempting to drag a new PlayerBody into the scene results in it being called "PlayerBody2"

This appears to be an issue in MovementProvider.gd in  _create_player_body_node() and how it sets the owner. Replacing
```player_body.set_owner(arvr_origin.owner)```
with:
```player_body.set_owner(get_tree().get_edited_scene_root())```
seems to fix the issue.

BastiaanOlij:
thanks!

PR #90, opened 2022-03-03
Show Script-generated PlayerBody in scene tree-view

Malcolmnixon:
This pull request fixes issue #87.


BastiaanOlij:
Makes sense, I guess my solution would only work if things aren't nested or something.

Issue #85, opened 2022-02-28
Jump movement should be able to be activated by swinging arms up


Henodude:
Right now, jumping can only be activated with a button press. I think an option should be added so that it can be activated by the player swinging their arms up.

Malcolmnixon:
A good idea! We were also thinking of detecting the player physically jumping by running a filter on the head/camera position.

PR #91, opened 2022-03-04
Added code to detect player jump movements

Malcolmnixon:
This change fixes issue #85 and consists of:
 - Moving the logic to perform the jump into PlayerBody
 - The one PlayerBody.player_jumped signal now gets called for all jumps (controllers, body, arms)
 - Added function to detect player jump (body and arms)

### DIFF
--- a/addons/godot-xr-tools/functions/Function_JumpDetect_movement.gd
+++ b/addons/godot-xr-tools/functions/Function_JumpDetect_movement.gd
@@ -1,0 +1,157 @@
+tool
+class_name Function_JumpDetect
+extends MovementProvider
+
+
+##
+## Movement Provider for Player Jump Detection
+##
+## @desc:
+##     This script can detect jumping based on either the players body jumping,
+##     or by the player swinging their arms up.
+##
+##     The player body jumping is detected by putting the cameras instantaneous
+##     Y velocity (in the tracking space) into a sliding-window averager. If the
+##     average Y velocity exceeds a threshold parameter then the player has
+##     jumped.
+##
+##     The player arms jumping is detected by putting both controllers instantaneous
+##     Y velocity (in the tracking space) into a sliding-window averager. If both
+##     average Y velocities exceed a threshold parameter then the player has
+##     jumped. 
+##
+
+
+## Movement provider order
+export var order := 20
+
+## Enable detecting of jump via body (through the camera)
+export var body_jump_enable := true
+
+## Only jump as high as the player (no ground physics)
+export var body_jump_player_only := false
+
+## Body jump detection threshold (M/S^2)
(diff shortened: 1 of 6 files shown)


addons/godot-xr-tools/functions/Function_JumpDetect_movement.gd
```diff
@@ -0,0 +1,154 @@
...
+
+	# Clamp the camera instantaneous velocity to +/- 2x the jump threshold
+	camera_vel = clamp(camera_vel, -2.0 * body_jump_threshold, 2.0 * body_jump_threshold)
+
+	# Get the averaged velocity
+	camera_vel = _camera_velocity.update(camera_vel)
+
+	# Detect a jump
+	if camera_vel >= body_jump_threshold:
+		player_body.request_jump()
```
BastiaanOlij, line 123: If the player physically jumps, should we then also add velocity to the jump or in this case just detect the jump and signal?

BastiaanOlij:
This looks good to me, worth adding a line to `versions.md` that we're adding this as its a new feature

addons/godot-xr-tools/functions/Function_JumpDetect_movement.gd
Malcolmnixon: I think it depends on the type of game - if someone's developing a basketball game then very few players are going to be able to pull off the Michael Jordan 4 foot vertical jump without a little "VR assistance". I was testing by jumping on one of my demo scenes jump pads to launch myself ~8 meters into the air.

As such, I think the best we can do is give the option to the developer. I'll add an exported property flag for 'physical jump only' which will report the jump rather than requesting it.

Malcolmnixon: I'm still thinking of suppressing the report if the player isn't on the ground, or the ground is too steep to jump according to the ground-physics settings. It doesn't seem to make sense to report the signal and potentially fire jump sound-effects if the player is in the middle of gliding when they physically jump.

Malcolmnixon: The PlayerBody.request_jump now supports an optional parameter (configurable option in JumpDetect) which suppresses the GroundPhysics velocity boost, so the jump only comes from the physical players jump.

BastiaanOlij: Owh cool, I can see a jungle gym sort of demo where you can jump around on bouncepads and grab onto the ceiling or something :)

BastiaanOlij:
Really annoying the version thing is causing merge conflicts git can't resolve... probably should do the other PR first, merge that, then rebase this one :)

PR #95, opened 2022-03-07
Object pickable grab feature without reparenting

goatchurchprime:
The current implementation moves the grabbed object with the controller by reparenting it under the controller.

This is inconvenient since it makes objects disappear from the node tree so it is impossible to write simple code that depends on objects staying where they were created.

We can get this the same effect by creating a RemoteTransform node under the controller and pointing it to the object that has been grabbed.  (It's not possible to reuse the same RemoteTransform over and over again since it does not allow the remote_path value to be set to null.)

### DIFF
--- a/addons/godot-xr-tools/images/teleport_arrow.png.import
+++ b/addons/godot-xr-tools/images/teleport_arrow.png.import
@@ -3,16 +3,17 @@
 importer="texture"
 type="StreamTexture"
 path.s3tc="res://.import/teleport_arrow.png-f1bd44b6f478277692b3fa29171b62d3.s3tc.stex"
+path.etc2="res://.import/teleport_arrow.png-f1bd44b6f478277692b3fa29171b62d3.etc2.stex"
 path.etc="res://.import/teleport_arrow.png-f1bd44b6f478277692b3fa29171b62d3.etc.stex"
 metadata={
-"imported_formats": [ "s3tc", "etc" ],
+"imported_formats": [ "s3tc", "etc2", "etc" ],
 "vram_texture": true
 }
 
 [deps]
 
 source_file="res://addons/godot-xr-tools/images/teleport_arrow.png"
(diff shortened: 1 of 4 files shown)


Malcolmnixon:
My testing shows the RemoteTransform works fine. You may wish to exclude the .import files as the changes were purely due to the gles3/gles2 project settings. The Viewport_2D_in_3D.tscn changes look to be a minor mismatch with the collision_layer settings between the top-level viewport and the StaticBody which the tool-script should have kept in sync.

BastiaanOlij:
Yup, agree with not having the .import files changes, those mostly relate to whether you open the project in 3.4.x or 3.5.

BastiaanOlij:
@goatchurchprime if you can change it so the .import files are not part of the PR I can merge this :)

BastiaanOlij:
@Malcolmnixon seeing you're doing the work on the ranged grab and making changes in this area at the moment, maybe you can combine @goatchurchprime change? Or maybe we should merge it and just undo the import file changes.

On the import file changes, I think we should try and stick to making sure the import file contents match what the current stable release of Godot expects. It's pretty unimpactful as it generally unintentional deviations from defaults now that Godot no longer saves the value of properties that haven't changed, but occasionally something unsupported by older versions sneaks in. Godot tents to ignore those but still....

Malcolmnixon:
I'm going to make the grab mechanic an enum as well - parenting, remote-transform, and possibly expanding to velocity tracking in a future pull request.

This is an area where things start to get interesting quickly because we're possibly going to need "grab-points" in the near future which might even have the ability to override the hand-pose (such as holding a gun) and if the object is velocity-tracked to prevent it from ghosting through walls, then we actually want the hand-model to stay with the object. As such we may expand this list of grab mechanics to include childinf the hand to the model, or remote-transforming the hand by the model.

Either way these features will take time to add, but I will include the chilling vs remote-transforming for now.

Malcolmnixon:
This capability has been added to the ranged-grab pull request. It now supports both Remote-Transform and Reparenting as the hold mechanic.

BastiaanOlij:
I shall close this than, thanks @Malcolmnixon and @goatchurchprime

PR #105, opened 2022-03-13
Modified jumping to use ground slope

### LINKED ISSUE #104: Jumping does not take into account the ground slope
Jumping should take into account the ground-slope. Doing so would allow for effects such as angled jump pads which can bounce the player up and to the side:
image

Malcolmnixon:
This pull request fixes #104 by using the ground surface-normal when applying the jump velocity.

Additionally this contains a minor fix to flight movement where the flight toggling button check is performed _process. This allows flying to be disabled even if a higher-priority exclusive movement (such as climbing) is being performed.


BastiaanOlij:
Thanks!

Issue #107, opened 2022-03-16
Pickable objects should have different highlight options


Malcolmnixon:
Currently packable objects only support showing highlighted state by setting the texture of a selected mesh to a predefined blue texture.

The highlighting should be expanded to support options such as showing or hiding nodes (such ss target rings).

A solution would be for pickables to emit signals on highlight state change, and having a set of highlighter nodes that can be added as children which auto-subscribe to the event.

Malcolmnixon:
The different highlighting options now seems to be working quite well in the demos.

PR #109, opened 2022-03-18
Enhanced pickable object highlighting

Malcolmnixon:
This pull request fixes issue #107 by adding numerous options for highlighting pickable objects. The change includes:
- Modified Object_pickable to emit a highlight_updated signal when the …highlight changes.
- Added a highlighter object which reacts to highlight_updated by changing mesh materials.
- Added a highlighter object which reacts to highlight_updated by showing/hiding itself (and children).
- Added a highlighter object which reacts to highlight_updated by showing/hiding a billboard target ring.


BastiaanOlij:
Very cool!

Issue #99, opened 2022-03-10
Removing scenes for functions and classes


BastiaanOlij:
Right now most of the features in XR Tools need to be added through subscenes. That makes a lot of sense as for many we need a handful of objects properly configured to make the function usable.

However we also define all classes with `class_name` which makes it possible to see them in the node tree and just add the node. If you do this, you get a node of the correct type, with the script attached, but it doesn't use the related scene.

Most of the scenes are really simple with just a handful of objects. It is fully possible to add the needed nodes in the `_ready` function of a script and remove the need to have the scenes.

Obviously this is something worth discussing whether we want to do this and to what type of classes to restrict this too. For instance, the scenes meant to be inherited should remain scenes, but we could fully remove the function scenes for instance.

Also this would be a serious breaking change and maybe something we should consider only for Godot 4.

Malcolmnixon:
Just talked with a user that's trying to position their player in the scene, and the player leaves a bunch of objects behind. These are all the helper objects under the movement functions (the wind sensor area), and as the MovementProvider isn't a Spatial node, the movement of the parent does nothing for the child.

Switching to script-generated helper objects would fix this problem - as would making MovementProvider extend from Spatial.

BastiaanOlij:
After discussing with Malcolm, we're slowly changing this over where we can, there is no point in keeping this issue open.

PR #110, opened 2022-03-19
Change wind movement so script can be used without scene

Malcolmnixon:
This pull request is related to #99 and makes the wind movement provider usable from the "Add Child Node" menu without needing the scene added.

It also contains modifications to conform better to the gdscript style guide, and contains a minor tuning-tweak to the JumpDetect default arm-jump threshold based on testing.


BastiaanOlij:
Cool! Works well as a template to start looking into the others too.

PR #111, opened 2022-03-20
Added ranged grabbing support to Function_Pickup

Malcolmnixon:
This pull request adds support for remote grabbing to Function_Pickup and was developed in collaboration with Sam Sarette (https://github.com/lunarcloud). Specifically these changes include:
 - Adding a ranged-grabbing area to Function_Pickup
 - Changing Function_Pickup so it can be added via the "Add Child Node" dialog by creating all child nodes dynamically
 - Adding a can_ranged_grab option to Object_Pickable

Note: The Function_Pickup script is essentially a complete rewrite in order to support the new features, and attempts to follow the gdscript style guide where possible.

### DIFF
--- a/addons/godot-xr-tools/CONTRIBUTORS.md
+++ b/addons/godot-xr-tools/CONTRIBUTORS.md
@@ -10,5 +10,6 @@ Other people who have helped out by submitting fixes, enhancements, etc are:
 - [Alessandro Schillaci](https://github.com/silverslade)
 - [jtank4](https://github.com/jtank4)
 - [Malcolm Nixon](https://github.com/malcolmnixon)
+- [Sam Sarette](https://github.com/lunarcloud)
 
 Want to be on this list? We would love your help.
--- a/addons/godot-xr-tools/VERSIONS.md
+++ b/addons/godot-xr-tools/VERSIONS.md
@@ -4,6 +4,7 @@
 - Improved responsiveness of snap-turning
 - Moved flight logic from Function_Direct_movement to Function_Flight_movement
 - Added option to disable player sliding on slopes
+- Added support for remote grabbing
 
 # 2.3.0
 - Added vignette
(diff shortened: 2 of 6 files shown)


BastiaanOlij (approved):

BastiaanOlij:
Cool stuff! Seeing you almost always make videos of features you add, make sure to add the link into the PR, makes it easier for to grasp exactly how things work: https://youtu.be/IKWM446kNSs

When we're working with objects up close, we work with an small area in which we grab anything so a player doesn't have to be precise and just snap whatever you're grabbing (though we should look into that avenue some day and see if we can switch between this play orientated grab style and a more precise grab style where  you have to be touching an object to grab it).

For remote grab I think that is a little jerky and I think we should add a tween in. Instead of snapping in place right away, we do the following (on XRToolsPickable.pick_up):
```
var original_transform = global_transform
original_parent.remove_child(self)
picked_up_by.add_child(self)

if by.picked_up_ranged:
  # position back where it was
  global_transform = original_transfrom
  
  # tween to new position
  var dest_transform = Transform()
  if center_pickup_on_node:
    dest_transform = center_pickup_on_node.global_transform.inverse() * global_transform
  
  # tween is a Tween node
  $Tween.interpolate_property(self, "transform", transform, dest_transform, 1, Tween.TRANS_LINEAR, Tween.EASE_IN_OUT)
  $Tween.start()
elif reset_transform_on_pickup:
  ...

```

Malcolmnixon:
The latest update does some fairly extensive changes to Object_pickable, including:
 - Support holding the object via RemoteTransform or Reparenting
 - Support ranged grabbing an object via Snap or Lerp
 - Cleanup to gdlint and godot style guide
 - Preparatory work to simplify adding snap-zones

I would like to consider the following changes:
 - Remove 'by_controller' as this doesn't appear to be used, and would be invalid for snap-zones
 - Rename 'original_mode' to be private

BastiaanOlij (approved):
Looks really good!

PR #118, opened 2022-04-08
Change compression import options on hand models to defaults so they'll load in HTML5

dsnopek:
The hand models had vertex compression (and only vertex compression) turned on, which (for some reason) prevents them from loading in an HTML5 export.

This PR changes the compression import options to their default values (which has Normal, Tangent, Color, TexUV, TexUV2 and Weights selected, but NOT vertex), allowing the hand models to load in HTML5.

_Related to https://github.com/godotengine/godot/issues/59949_


BastiaanOlij:
Funky, thanks for fixing this @dsnopek  :)

Issue #115, opened 2022-04-05
Feature Request: Separate Turning from Direct Movement Function


teddybear082:
There is a control convention where direct or teleport movement is located on the dominant hand along with strafing (which present XR-tools supports) but then the off hand just has snap or smooth turning and can utilize the Y axis up /down for non-movement in-game controls, like quick switching between weapons, or jumping and crouching.  This is fairly trivial to hack with the current tools by copying the direct movement scene and script and scraping out anything related to direct movement but it would be good to officially have its own module so people could mix and match, similar to what has now been done with separating out flight movement from direct movement.

My hacked files are here for turn only: https://drive.google.com/drive/folders/1osoVU7tWHyNJwMDtK8jpxOt9aVN4Rsgb?usp=sharing

Malcolmnixon:
I've got a pull request implementing this change and also added it to https://github.com/Malcolmnixon/godot-xr-tools-demo.

teddybear082:
I think we can close this issue now right?

Malcolmnixon:
I think this one has had enough testing, so yes we can close it.

PR #121, opened 2022-04-20
Added Function_Turn_movement provider

Malcolmnixon:
This pull request extracts the turn logic from Function_Direct_movement and moves it into a Function_Turn_movement provider. The benefits of this are:
 * Simpler and more modular code
 * It's possible to add movement and turning separately
 * Movement can be disabled without disabling turning

The only down-side is that existing code will now have to update the Function_Direct_movement settings, and possibly add a Function_Turn_movement to keep existing player movement behavior.

This pull request implements feature request #115

### DIFF
--- a/addons/godot-xr-tools/VERSIONS.md
+++ b/addons/godot-xr-tools/VERSIONS.md
@@ -5,6 +5,7 @@
 - Moved flight logic from Function_Direct_movement to Function_Flight_movement
 - Added option to disable player sliding on slopes
 - Added support for remote grabbing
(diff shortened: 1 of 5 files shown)


BastiaanOlij:
Sorry for the late response Malcolm, looks good, my only concern, as you already mention, is that this is a breaking change for people using previous versions of the library. Might not be avoidable, I do like this split.

Malcolmnixon:
I'll push up an update this evening with the VERSIONS.md updated to mention the change.

BastiaanOlij (approved):

Issue #129, opened 2022-05-16
Feature Request - Snap Zones


Malcolmnixon:
Pickable objects should support being "snapped" into snap-zones. These snap-zones should support being fixed to the environment, or put on the player as pockets.

teddybear082:
I liked your video on this! I closed my "holster" issue accordingly.  Not sure if it is in your code but would be good to have a check box in the editor to say whether the object should be hidden or not when placed in the snap zone, e.g., difference between using the snap zone on a "back pack" object that hides it, or using the snap zone on a shoulder holster where you want other players to continue to see the object on your back, like a sword.

Malcolmnixon:
Tested in 2.4 release.

PR #130, opened 2022-05-16
Added support for snap-zones

Malcolmnixon:
Added support for snap-zones. This pull request implements feature request #129 and contains the following changes:
* pick_up() can now forward the pickup to a different object. This lets picking up a snap-zone instead grab the object in the snap-zone.
* Object_pickable can now accept/refuse being picked up by certain objects. This can be useful for preventing a wrist-pocket snap-zone from being grabbed by the Function_pickup on the hand.
* Function_pickup more cleanly handles its controller being disabled. This prevents some "det==0" warnings.
* Snap_Zone script and scene added.


BastiaanOlij (approved):
Looks good, there might be one or two small breakage issues with scripts that override can_pick_up in the old way but I think we can accept that risk as its easily fixed.

Issue #144, opened 2022-06-14
Feature Request - fall detection


Malcolmnixon:
Add a module for detecting when the player takes fall (or collision damage).

Malcolmnixon:
Fall detection example module present and working in demo projects.

PR #145, opened 2022-06-14
Added fall damage detection module

Malcolmnixon:
This pull request implements feature #144 for fall detection. The implementation works by detecting instantaneous accelerations above a given threshold. There are options for:
 * Suppress acceleration/damage caused by the player being launched into the air - for example by jump pads
 * Detect only damage caused by colliding with the ground, as opposed to colliding with a wall


BastiaanOlij:
Hmmm, I'm not sure about this one, it's a bit strange implementing specific game logic into a generic toolkit. I need to think about this and whether there is a more generic approach, this has the danger of implementing all sorts of specifics as people want to detect other forms of impacts.

Malcolmnixon:
I moved the fall-detector into an "examples" folder and put an extensive comment explaining that it was an example and that games would most likely need to customize it.

Issue #143, opened 2022-06-10
Feature Request - fix climbing to allow pulling onto the top of objects


teddybear082:
Here's my version.  https://github.com/teddybear082/Godot_XR_Tools_Climb_Fix  

I know you all can do better than me!

I also know new features are paused until all bugs are found but just recording this since I know it is a feature you want to put in at some point so will need an issue for the pull request anyway.

teddybear082:
Confirmed this works!! Yay!

Issue #116, opened 2022-04-05
Feature Request - Crouch Function


teddybear082:
Presently it appears room scale crouching is supported but button crouch is not an official function.

Malcolmnixon:
This goes quite nicely with some other features we've been thinking of which control the players body, including:
 - height calibration (for tall players and people in wheel chairs)
 - shrinking the player height while climbing to allow for pulling the body on to the top of buildings.

teddybear082:
@Malcolmnixon I have an alpha version of getting climbing to the top of platforms and buildings working by adding / replacing this code to the climbing function after the code where you have the average offset for left and right hands combined.  Does this approach make sense to you?

https://pastebin.com/PFT5ESrz

teddybear082:
@Malcolmnixon is your controller suffering from drift or something else? If its broken due to drift I have found that compressed air and electronic contact cleaner (mine is CRC Mass Air Flow Sensor Cleaner) work wonders.

Malcolmnixon:
Could you check out the crouch functionality in the latest https://github.com/Malcolmnixon/godot-xr-tools-demo and let me know if it works for you?

teddybear082:
This works!! Nice work! I couldn't figure how to do this no matter how hard I thought about it and tried.

PR #146, opened 2022-06-15
Added player height control to support mounting climbed objects and crouching

Malcolmnixon:
This pull request implements feature requests #143 and #116 with the help of @teddybear082. The changes include:
 - Added advanced player height control
 - Modified climbing to collapse player to a sphere to allow mounting climbed objects
 - Added crouch movement provider

The PlayerBody now supports:
 - Player height offset to allow future height calibration (for seated or disabled players)
 - Minimum and maximum allowed player height (as measured by head position)
 - Overriding player height by movement providers (for crouching or climb-mounting)

The math for placing the player body under the camera had to be modified extensively - the original logic would *always* put the player body on the ARVROrigin floor, whereas the new logic puts the player body under/around the head. This change results in the head/camera always being in the correct position in relationship to the body.

### DIFF
--- a/addons/godot-xr-tools/functions/Function_Climb_movement.gd
+++ b/addons/godot-xr-tools/functions/Function_Climb_movement.gd
@@ -122,11 +122,13 @@ func _set_climbing(active: bool, player_body: PlayerBody) -> void:
 	if is_active:
 		_distances.clear()
 		_deltas.clear()
+		player_body.override_player_height(self, 0.0)
 		emit_signal("player_climb_start")
 	else:
 		var velocity := _average_velocity()
 		var dir_forward = -(player_body.camera_node.global_transform.basis.z * HORIZONTAL).normalized()
 		player_body.velocity = (velocity * fling_multiplier) + (dir_forward * forward_push)
+		player_body.override_player_height(self)
 		emit_signal("player_climb_end")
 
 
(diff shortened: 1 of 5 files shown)


addons/godot-xr-tools/functions/Function_Climb_movement.gd
```diff
@@ -122,11 +122,13 @@ func _set_climbing(active: bool, player_body: PlayerBody) -> void:
 	if is_active:
 		_distances.clear()
 		_deltas.clear()
+		player_body.override_player_height("climb", 0.0)
```
BastiaanOlij, line 125: here you hardcode a string "climb" while in crouch you use self. We should chose either one or the other and be consistent

BastiaanOlij (approved):
This is a really nice enhancement :)

addons/godot-xr-tools/functions/Function_Climb_movement.gd
Malcolmnixon: I made the override_player_height take a variant so it could be any type of key. Originally I was just using strings, but then I found that I wanted the left and right controller crouch buttons to both submit separate crouch requests which is why I made it take a variant and just used the nodes. You are correct however that in this case using 'self' may actually be cleaner - in debugging it would then be possible to find out which object is requesting the height adjustment and actually point back to the node. 

I'll push up the change.

Malcolmnixon:
It should be ready to go in now.

Issue #156, opened 2022-06-30
Feature request: object teleport block


0968128:
Is there already an option to make it impossible to teleport onto certain objects? Not just place it on a different layer, so the teleporter goes through, but an option to actually block the ray and make it red.

Malcolmnixon:
You may want to check out the pull request and see if it achieves what you're after. With this the Function_Teleport now has a Valid Teleport Mask which specifies the object layers where the teleport is permitted.
![image](https://user-images.githubusercontent.com/1863707/176801174-b9a75bef-a799-4703-ab47-0261a7c441f9.png)
The default value of this new property is "everything", so existing projects should not be broken.

Malcolmnixon:
This was tested in the teleport demo of https://github.com/Malcolmnixon/godot-xr-tools-demo by preventing teleporting onto the rotating carousel.

PR #157, opened 2022-07-01
Added valid-layer checking to teleport movement

Malcolmnixon:
This change implements feature request #156 by checking the final target object and rejecting it if the objects layer isn't in our valid_teleport_mask layer-mask. This allows for objects that block the teleport-arc, but are themselves not valid teleport targets.


BastiaanOlij:
@Malcolmnixon wasn't I using the collision mask of the kinematic body for that already?

BastiaanOlij:
Makes sense :)

PR #159, opened 2022-07-10
Modified hand meshes to be pre-scaled so the scenes can be 1:1 scaled

Malcolmnixon:
This pull request modifies the scales on the following hand meshes:
 - LeftHand.blend
 - LeftHand.glb
 - RightHand.glb (not requred, as RightHand.tscn uses LeftHand.glb flipped in the X axis)

This simply applied the existing 0.068 scale present in LeftHand.blend (Set in the object transform, but not applied); and then applied another 0.5 scale present in the LeftHand.tscn and RightHand.tscn scenes. The scales were then set to 1:1 in the scenes.

The result is that the "LeftHand\LeftHand" node which used to have a 0.5, 0.5, 0.5 scale is now 1,1,1:
image

Additionally the "LeftHand\LeftHand\Armature_Left" node which used to have a 0.068, 0.068, 0.068 scale is also 1,1,1:
image


BastiaanOlij (approved):
This is something we need to be mindful of in the future. It is much better to have assets imported at their correct scales instead of applying scales in the engine

PR #168, opened 2022-07-26
Restructure plugin to include a baseline demo

BastiaanOlij:
This is our first step to restructuring our repo. This just introduces the base files for our demo. I'll likely merge this as is so I can start working on the CI that creates our deliverables, while Malcolm starts applying a lot of the refactoring.


BastiaanOlij:
Ok, this is pretty much ready to be merged. On a PR it will now do a build of both the plugin by itself, and the full demo including embedding the OpenXR library.

When tagging it prepares a full release with those two entities.
Here is an example of what that looks like: https://github.com/BastiaanOlij/godot-xr-tools/releases/tag/3.0.0-test

Note that when you examine the zip files they have a root directory, this may seem a bit duplicitous but it is required to publish these zip files on the asset library.

.github/workflows/build-on-push.yml
```diff
@@ -0,0 +1,64 @@
...
+          mkdir plugin/godot-xr-tools/addons
+          cp -r demo/godot-xr-tools/addons/godot-xr-tools plugin/godot-xr-tools/addons
+          cp demo/godot-xr-tools/LICENSE plugin/godot-xr-tools
+          cp demo/godot-xr-tools/CONTRIBUTORS.md plugin/godot-xr-tools
+          cp demo/godot-xr-tools/VERSIONS.md plugin/godot-xr-tools
+          rm -rf demo/godot-xr-tools/.git
+          rm -rf demo/godot-xr-tools/.github
+      - name: Add OpenXR Plugin
+        run: |
+          wget --header "Authorization: token ${{ secrets.GITHUB_TOKEN }}" "https://github.com/GodotVR/godot_openxr/releases/download/1.3.0/godot-openxr.zip"
```
BastiaanOlij, line 33: For now we'll need to hardcode the version of the OpenXR plugin we embed in our demo asset.

Malcolmnixon: I think this works fine for now. I was thinking it could be a build-job variable accessed by the yaml; however I can't find how to set variables through the web; and even if it worked, we might want to have different versions specified for different branches - something embedding in the yaml is capable of achieving.

It does mean we have to update the yaml when the repo wants to jump forwards a rev; but that seems a small price to pay for having the explicit control in branches.

Malcolmnixon (approved):
This looks perfect!

PR #169, opened 2022-07-27
Create staging and scene loading setup

BastiaanOlij:
This introduces a staging and background loading setup for our demos.

At the core we have a new starting scene called `staging.tscn` which contains a loading screen and all the logic to load in supporting scenes. 

Loading the scenes that form our actual levels/demos is now done using a background loader. While a scene is loading our loading screen is shown. Before we actually switch to our loaded scene the user is asked to hold the trigger button. We may make this an optional step that is only active if we detect the player isn't wearing their headset (yet) by checking the position of the HMD.

At startup we will default to loading `res://scenes/main_menu/main_menu_level.tscn`, this scene will allow us to load different parts of the demo.

Each scene that can be loaded should inherit from `res://scenes/scene_base.tscn` to ensure basic logic is in place.
These scenes can emit the following two signals:
- `exit_to_main_menu` will return the user to the main menu level
- `load_scene(scene)` will load the given scene and switch to it.

Note that our base scene has an ARVROrigin+ARVRCamera setup that can be further build out. The loading system already ensures that the main origin in the staging scene gets disabled.

Finally this PR introduces a new helper scene for XR tools called `XRToolsHoldButton`, this scene when enabled will detect the user pressing a given button on any controller and will emit a signal when the button is held for a preset amount of time after which the button disables itself.


Malcolmnixon (approved):
It looks good and works fine!

Issue #170, opened 2022-07-30
Add bouncy property to physics


GammaGames:
To make the player ricochet off stuff, because that's a good idea in VR

Malcolmnixon:
@GammaGames could you check out the branch and let me know if it's doing what you want. You can also play with the latest https://github.com/Malcolmnixon/godot-xr-tools-demo which has a few trampolines (purple plates) scattered around - the ones beside the climb-building are sloped so you can test angled-collision bounces.

Note: this only works if the movement is being performed by the PlayerBody. It won't be applied if an exclusive movement-provider (such as the grappling hook) has taken over physics.

GammaGames:
You're fast, I'll try to check it out tomorrow! Thanks!

GammaGames:
I played around with the demo, looks good!

PR #171, opened 2022-07-31
Added support for bouncing to the player body

Malcolmnixon:
This pull request implements feature request #170 for bounciness as follows:
 - GroundPhysicsSettings now contains bounciness and bounce threshold
 - PlayerBody detects collisions during move_and_slide, and can bounce the player appropriately

Additionally GroundPhysics (the helper node to attach a GroundPhysicsSettings to collider nodes) has a new helper get_physics static function to help access the physics. This is to help simplify duplicate complex code in the PlayerBody class.

The approach to bouncing is as follows:
 - Detect when the player body performs a move_and_slide that contacts a collider
 - Get the bounciness parameters applicable to the body (or inherited from the player)
 - Check the magnitude of the player velocity (pre-collision) into the collider
 - Check if the collision-magnitude is larger than the bounce-threshold
 - If so, reflect the original (pre-collision) velocity on the plane defined by the normal vector, and apply to the body

This is attempting to implement more physics-like effects for the player body. At some point it may be worth investigating using a RigidBody and some serious _integrate_forces code; although that might be better done as a Godot 4 physics extension.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -4,6 +4,7 @@
 # 2.6.0
 - Fixed enforcement of direct-movement maximum speed
 - Added editor icons for all nodes
+- Added collision bouncing to PlayerBody
 
 # 2.5.0
 - Added advanced player height control
--- a/addons/godot-xr-tools/assets/PlayerBody.gd
+++ b/addons/godot-xr-tools/assets/PlayerBody.gd
@@ -24,6 +24,9 @@ extends Node
 ## Signal emitted when the player jumps
 signal player_jumped()
 
+## Signal emitted when the player bounces
+signal player_bounced(collider, magnitude)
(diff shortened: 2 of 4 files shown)


BastiaanOlij:
We only just changed so this is more for the future, we should attempt to do changes in master first and then either cherry pick or redo the work for 2.x afterwards (cherry pick will likely not be feasible for long seeing all the reorganisation we're planning).

BastiaanOlij (approved):

BastiaanOlij:
@Malcolmnixon don't forget the rebase on this one so it can be merged :)

Malcolmnixon:
Just rebased onto master and re-tested it in my demo scene with the trampolines.

PR #176, opened 2022-08-02
Added initial demo scene with teleporters for scene-transitions.

Malcolmnixon:
Unregister signals from staging before fading to prevent processing duplicate scene-switch events.

Added meshes and supporting scenes/materials/shaders/scripts for mound, ramps, and teleport objects.
Added basic movement demo scene.
Added teleport from main menu level to basic movement demo scene.


assets/meshes/teleport/teleport.gd
```diff
@@ -0,0 +1,50 @@
...
+		return
+
+	# Teleport
+	if scene:
+		_scene_base.emit_signal("load_scene", scene.resource_path)
+	else:
+		_scene_base.emit_signal("exit_to_main_menu")
+
+func _set_title(value):
+	print("Setting title to ", value)
```
lunarcloud, line 41: Remove your print statements @Malcolmnixon

assets/meshes/teleport/teleport.gd
```diff
@@ -0,0 +1,50 @@
...
+	title = value
+	if is_inside_tree():
+		_update_title()
+
+func _update_title():
+	print("Updating title to ", title)
```
lunarcloud, line 47: Here too

lunarcloud (changes requested):
You missed these

assets/meshes/teleport/teleport.gd
Malcolmnixon: Good catch - left that debug print statement in there.

assets/meshes/teleport/teleport.gd
Malcolmnixon: Fixed as well.

Malcolmnixon:
I just pushed another update which adds a second demo scene for teleporting. This one also has the snap-turn provider in there which has two side-effects:
 1. The player has a body, and so can trigger the teleport-pads to leave
 2. The player has a body, and so can be affected by slopes and gravity

BastiaanOlij:
Finally had a change to try this out. Very cool! I really like the teleporters!

BastiaanOlij (approved):

PR #178, opened 2022-08-11
Added climbing and grappling scenes to the demo project

Malcolmnixon:
Added climbing demo scene.
Added cloudy panorama sky from ambientcg.com.
Cleaned up lighting in scenes.
Minor cleanups for demo-scene instructions.


Malcolmnixon:
Added wind areas and the wind movement provider to the climbing/gliding demo.

Malcolmnixon:
Added grappling scene to the pull request.

BastiaanOlij:
Had a bit of a play around with the new test scenes, love the grappling, nice and simple :)

I did get stuck in the gravity lifts in the climbing demo. Not sure why, we might need to add an option to reset the player to the start position.

Malcolmnixon:
It's possible (but not obvious) to get out using a t-pose glide when the player is bouncing at the top of the wind column to slowly move forwards.

A return-to-main-menu button would be quite useful, rather than running all the way back to a teleporter.

BastiaanOlij (approved):

BastiaanOlij:
> It's possible (but not obvious) to get out using a t-pose glide when the player is bouncing at the top of the wind column to slowly move forwards.
> 
> A return-to-main-menu button would be quite useful, rather than running all the way back to a teleporter.

New players won't realise the T-pose is an option, I bet many won't even read the billboard at the start :) So having an easy way to reset or bring up a menu makes sense to me. We could do that as a separate PR though

PR #181, opened 2022-09-02
Merged interactable scripts from godot-xr-tools-interactables project

Malcolmnixon:
Added interactable scripts.
Added interactable demo scene and objects.


BastiaanOlij (approved):
Very cool stuff!

PR #182, opened 2022-09-05
Added pointer demo scene

Malcolmnixon:
Added pointer demo including on-screen keyboard and color-changing cubes.


BastiaanOlij (approved):
Works a treat thanks!

PR #184, opened 2022-09-15
Change GroundPhysicsFlags enum to use binary

Henodude:
This makes it a bit more clear what the numbers are for. Also added a trailing comma after the last item, for cleaner diffs if more flags are added.


Malcolmnixon:
I like it. I tend to use hex more often than not due to less characters, but in this case with the small number of options, I think the binary works best.

Malcolmnixon (approved):

PR #185, opened 2022-09-15
Added picakable demo with snap-zones

Malcolmnixon:
Added pickable demo.
Named physics layers in project.


BastiaanOlij:
Hmmm, I'm not sure I like having the different objects fully embedded in the scene. I would have expected inheriting `Object_pickable.tscn` and then configuring the ball in there and saving that as `ball.tscn`. Then that scene can be included a number of times to show re-usability. 

Seeing the visual aspects with these type of objects the base scene comes in very handy.

BastiaanOlij (approved):
Brilliant!

Issue #188, opened 2022-09-18
Scripts should be easier to upgrade using godot 3 to 4 upgrader


Malcolmnixon:
Godot-xr-tools uses numerous scripting approaches which don't work well with the godot 3 to 4 upgrader. These include:
 * Rename PlayerBody move_and_slide() function to move_body() as the name-collision confuses the upgrader
 * NodePath default argument values can no-longer be empty strings - instead replace with NodePath("")
 * GDScript 2 doesn't do as good a job with type-inference so prefer explicit types
 * Typed-exports don't work, so replace "export (SomeType) var x" with "export var x: SomeType"

Malcolmnixon:
Changes to aid in auto-conversion have been merged, and additionally the scripts have all been aligned between the 3.x and 4.x branches.

PR #189, opened 2022-09-18
Implemented script changes to help the godot 3 to 4 upgrader.

Malcolmnixon:
This pull request implements #188 by making the scripts easer for the upgrader to process.

### DIFF
--- a/addons/godot-xr-tools/assets/PlayerBody.gd
+++ b/addons/godot-xr-tools/assets/PlayerBody.gd
@@ -57,7 +57,7 @@ export var gravity := -9.8
 export var push_rigid_bodies := true
 
 ## GroundPhysicsSettings to apply - can only be typed in Godot 4+
-export (Resource) var physics = null setget set_physics
+export var physics: Resource = null setget set_physics
 
 # Set our collision layer
 export (int, LAYERS_3D_PHYSICS) var collision_layer = 1 << 19 setget set_collision_layer
@@ -275,8 +275,8 @@ func request_jump(var skip_jump_velocity := false):
 	emit_signal("player_jumped")
 	_jump_cooldown = 4
 
-# Perform a move_and_slide on the kinematic node
-func move_and_slide(var velocity: Vector3) -> Vector3:
+# Move the players body
+func move_body(var velocity: Vector3) -> Vector3:
 	return kinematic_node.move_and_slide(velocity, Vector3.UP, false, 4, 0.785398, push_rigid_bodies)
 
 # Set or clear a named height override
@@ -294,7 +294,7 @@ func override_player_height(key, value: float = -1.0):
 # This method updates the body to match the player position
 func _update_body_under_camera():
 	# Calculate the player height based on the camera position in the origin and the calibration
-	var player_height := clamp(
+	var player_height: float = clamp(
 		camera_node.transform.origin.y + player_head_height + player_height_offset,
 		player_height_min * ARVRServer.world_scale,
 		player_height_max * ARVRServer.world_scale)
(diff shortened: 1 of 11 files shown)


addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -57,7 +57,7 @@ export var gravity := -9.8
 export var push_rigid_bodies := true
 
 ## GroundPhysicsSettings to apply - can only be typed in Godot 4+
-export (Resource) var physics = null setget set_physics
+export var physics: Resource = null setget set_physics
```
BastiaanOlij, line 60: I didn't know this notation already worked in Godot 3, cool :)

addons/godot-xr-tools/functions/Function_Teleport.gd
```diff
@@ -40,7 +40,7 @@ export (int, LAYERS_3D_PHYSICS) var valid_teleport_mask = ~0
 # export (int, LAYERS_3D_PHYSICS) var collision_mask = 1
 
 # We don't know the name of the camera node...
-export (NodePath) var camera = null
+export var camera: NodePath = NodePath("")
 
 onready var ws = ARVRServer.world_scale
 var origin_node = null
```
BastiaanOlij, line 46: I don't know if the converter does this automatically but I remember when I did the manual conversion it really didn't like when variables meant to be used as nodes weren't typed. Maybe we should pre-empt that here and as a principle start typing these types of variables?

So in this case: `var origin_node : XROrigin`

Also the `= null` gave issues for me, it is enough to just type the node and it will default to empty correctly.

addons/godot-xr-tools/assets/PlayerBody.gd
```diff
@@ -483,7 +483,7 @@ func _get_configuration_warning():
 	return ""
 
 ## Find the Player Body from a player node and an optional path
-static func get_player_body(node: Node, var path: NodePath = "") -> PlayerBody:
+static func get_player_body(node: Node, var path := NodePath("")) -> PlayerBody:
```
BastiaanOlij, line 486: Nitpicking here but should we do:
```
static func get_player_body(node: Node, var path : NodePath = NodePath("")) -> PlayerBody:
```
for consistency? it's a little more verbose but it seems that Godot 4 really does prefer typed variables over untyped even if in this case it can easily infer it.

addons/godot-xr-tools/objects/Viewport_2D_in_3D.gd
```diff
@@ -8,7 +8,7 @@ export var enabled = true setget set_enabled
 export var screen_size = Vector2(3.0, 2.0) setget set_screen_size
 export var viewport_size = Vector2(300.0, 200.0) setget set_viewport_size
 export var transparent = true setget set_transparent
-export (PackedScene) var scene = null setget set_scene, get_scene
+export var scene: PackedScene = null setget set_scene, get_scene
```
BastiaanOlij, line 11: I wonder if this should be ` = PackedScene()` instead of ` = null` or leaving the assignment out all together as the default is an empty object.

BastiaanOlij (approved):
This is an exciting start to getting things ported to Godot 4 more easily.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: I just re-checked and it does select the resource picker in Godot 3.4.5. This is one where it'll work in 4, but we'll still probably want to upgrade it to the typed-resource anyway.

addons/godot-xr-tools/assets/PlayerBody.gd
Malcolmnixon: Agreed, I switched them all over to the long-form style.

addons/godot-xr-tools/functions/Function_Teleport.gd
Malcolmnixon: I removed the null assignment; but I don't think node-path typing is possible in 3.x and instead requires the new 4.x format "@export_node_path(Button, TouchScreenButton) var some_button"

addons/godot-xr-tools/objects/Viewport_2D_in_3D.gd
Malcolmnixon: I removed the null assignment as gdscript is going to do a better job of picking a default/empty value.

PR #192, opened 2022-09-28
Minimal changes to make code functional with Godot 4.0 beta 1

Malcolmnixon:
This pull request makes the older godot-xr-tools (based on 2.x) work with Godot 4.0 beta 1. The following has been tested:
 * Hands
 * Player body
 * Direct movement
 * Teleport movement
 * Flying
 * Gliding
 * Jumping
 * Pickup
 * Pickable Objects
 * Climbing
 * Climbable Objects

The Vignette has had to have its tool-script disabled to prevent locking up Godot 4 on startup.


BastiaanOlij (approved):
Looks good!

BastiaanOlij:
I'm not sure yet why vignette doesn't work but it also doesn't really need to be a tool script, I need to find some time to do further testing but if all else fails we can remove the tool feature.

PR #194, opened 2022-10-06
Merged in all master features and bug fixes.

Malcolmnixon:
This pull request merges in all master features into the 4.0-dev branch. Many of them have been tested including:
 - New 1:1 scaled hands and hand collisions
 - Interactables
 - New movement providers (crouch, turn, grapple)
 - Remote grabbing and highlights

Some need a little more work including Viewport2Din3D interactions, and the vignette is still not working.


Malcolmnixon:
Forgot the fall damage detector from the examples folder - just added

addons/godot-xr-tools/examples/fall_damage.gd
```diff
@@ -0,0 +1,86 @@
+@tool
```
BastiaanOlij, line 1: Should we move this into the demo? (possibly as a separate thing seeing we already did this in this way for Godot 3)

BastiaanOlij (approved):
I'd say, lets go for it, we're not going to find what we missed until we start playing around with the demo

PR #197, opened 2022-10-10
Added basic port of the demo project.

Malcolmnixon:
This pull request brings in the demo project and does the minimal amount of porting necessary for the demo to run. There are numerous issues with the demo at this point:
 * Fade in/out not working
 * Panorama sky distortions
 * VR keyboard text input not being sent to 2D screens
 * Snap zones not snapping


BastiaanOlij:
The Panorama Sky I think is a bug in Godot itself, the rest we'll have to tackle one by one. I'll have a look into the fade in/out issue, there are a few things that don't work in converting shaders from Godot 3 to 4...

VERSIONS.md
```diff
@@ -0,0 +1,89 @@
+# 3.0.0 Alpha
```
BastiaanOlij, line 1: Pretty sure we're missing entries here or have never updated them in master, you added a lot more to the demo than `included demo` justifies :)

We should also add a

```
# 4.0.0
- Conversion to Godot 4
```

BastiaanOlij (approved):
As you suggested, lets merge this and work on resolving the left over issues one at a time. 

Then we need to have a discussion on whether we're going to keep 4.0-dev, or make this master and have a 3.x branch to back port things too?

PR #198, opened 2022-10-11
Script cleanup

Malcolmnixon:
This pull request back-ports numerous improvements from the 4.0-dev branch, including:
 * Adding type hints where appropriate and in preference of type-inference
 * Adding a few returns if Editor.editor_hint
 * Make spacing match 4.0-dev to simplify future merges
 * Remove unwanted var in function parameters

### DIFF
--- a/addons/godot-xr-tools/assets/hand.gd
+++ b/addons/godot-xr-tools/assets/hand.gd
@@ -20,11 +20,11 @@ signal hand_scale_changed(scale)
 
 
 # Last world scale (for scaling hands)
-var _last_world_scale := 1.0
+var _last_world_scale : float = 1.0
 
 
 # Capture the initial transform
-onready var _transform := transform
+onready var _transform : Transform = transform
 
 
 # Called every frame. 'delta' is the elapsed time since the previous frame.
--- a/addons/godot-xr-tools/functions/movement_jump.gd
+++ b/addons/godot-xr-tools/functions/movement_jump.gd
@@ -2,12 +2,13 @@ tool
 class_name XRToolsMovementJump
 extends XRToolsMovementProvider
 
+
 ##
 ## Movement Provider for Jumping
 ##
 ## @desc:
 ##     This script provides jumping mechanics for the player. This script works
-##     with the PlayerBody attached to the players ARVROrigin.
+##     with the XRToolsPlayerBody attached to the players ARVROrigin.
 ##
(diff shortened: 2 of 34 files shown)


addons/godot-xr-tools/assets/hand.gd
```diff
@@ -20,15 +20,19 @@ signal hand_scale_changed(scale)
...
 
 
 # Capture the initial transform
-onready var _transform := transform
+onready var _transform : Transform = transform
 
 
 # Called every frame. 'delta' is the elapsed time since the previous frame.
 func _process(_delta: float) -> void:
+	# Do not process if in the editor
```
BastiaanOlij, line 32: `_process` only runs in the editor for tool scripts. I guess we could add this for clarity just in case someone decides to turn this into a tool script in the future but otherwise it's overkill.

addons/godot-xr-tools/functions/movement_jump.gd
```diff
@@ -2,12 +2,13 @@ tool
...
 extends XRToolsMovementProvider
 
+
 ##
 ## Movement Provider for Jumping
 ##
 ## @desc:
 ##     This script provides jumping mechanics for the player. This script works
-##     with the PlayerBody attached to the players ARVROrigin.
+##     with the player body attached to the players ARVROrigin.
```
BastiaanOlij, line 11: If we are referring to something that is a class, I think it makes sense to use the class name here, so `XRToolsPlayerBody` I guess?

addons/godot-xr-tools/interactables/interactable_slider.gd
```diff
@@ -96,13 +96,13 @@ func move_slider(var position: float) -> void:
 
 
 # Handle release of slider
-func _on_slider_released(var _interactable):
+func _on_slider_released(_interactable):
```
BastiaanOlij, line 99: No type here?

BastiaanOlij (approved):
Looks good :)

addons/godot-xr-tools/assets/hand.gd
Malcolmnixon: I found a few scripts with editor_hint checks that weren't tool scripts and cleaned those up.

addons/godot-xr-tools/interactables/interactable_slider.gd
Malcolmnixon: Added typing for the interactable being release (always self for this self-subscribing handler)

PR #200, opened 2022-10-12
Normalization of 3.x and 4.x scripts

Malcolmnixon:
This normalizes the 3.x and 4.x scripts to take into account the latest 3.x changes and the set-function syntax.


BastiaanOlij (approved):
Looks good!

Issue #208, opened 2022-10-14
Scrolling titles on teleport get blockier after a few scene changes


goatchurchprime:
I'd say it was something wrong with the mipmaps, but this is some kind of shader, so I have no idea what could be up.
![image](https://user-images.githubusercontent.com/677254/195909685-81b0bacc-5dac-49c0-8f38-9eba6160678d.png)

Malcolmnixon:
Fascinating. I usually run in and out of all scenes when doing a full test of changes and I don't see that quantization of the X axis.

Could it be time-related rather than the number of scene transitions? It might be an effect where TIME gets numerically large in the shader and so fine detail is lost. If so then I might need to add a modulo In at some point earlier in the time path.

Malcolmnixon:
Could you check out the pull request and verify whether it fixes your issue. I was only able to artificially induce it on my system.

PR #211, opened 2022-10-14
Fix quantization of teleport title image

Malcolmnixon:
This pull request fixes issue #208 which appears to occur due to the TIME in the shader getting numerically large (up to 3600.0) resulting in loss of precision in the texture UV coordinates.

This change introduces the Remainder (modulo) operation immediately after the TIME input block so the time remains in the interval [0..5)
image


BastiaanOlij (approved):
Time can be an annoying variable to work with for sure, seems like a sound fix to me.

Issue #199, opened 2022-10-11
Viewport2Din3D mouse mimic forgets to set the pressure


goatchurchprime:
I was working on bringing a 2D drawing addon into VR which uses the `event.pressure` value to set the darkness of the line.
https://github.com/mbrlabs/Lorien/issues/207

It appears that in normal 2D apps which use `func _input(event)` the incoming `InputEventMouseMotion` has its `pressure` value set to 1 if the left mouse button is down, and zero if it is up.
    
Therefore, to make the `godot-xr-tools/objects/Viewport_2D_in_3D_body.tscn` consistent with mouse events, we should include the line:

> event.set_pressure(mouse_mask)

in the `_on_pointer_moved()` function below the `# Let's mimic a mouse` comment somewhere about here:
https://github.com/GodotVR/godot-xr-tools/blob/master/addons/godot-xr-tools/objects/viewport_2d_in_3d_body.gd#L49

Malcolmnixon:
The pressure field was added in 3.2 so it should be supported on all Godot versions we support.

Malcolmnixon:
@goatchurchprime Could you confirm the change works for you, and if so close this issue?

goatchurchprime:
Yes that works.  It came in when I updated my version of the toolkit that I have symlinked from my addons directory and checked out the 3.0.0 tag.  (I was wondering why my hand pointer was causing so much problems suddenly).

PR #216, opened 2022-10-22
Added setting the mouse movement pressure

Malcolmnixon:
Added setting the mouse movement pressure when simulating the viewpoint pointer movement.

This fixes issue #199.


BastiaanOlij (approved):
Owh seems like a nobrainer, I wonder when `set_pressure` was added but seeing we're targeting 3.5 for XRTools 3.0.0 anyway we should be able to rely on it existing.

Issue #217, opened 2022-10-23
Auto Pointer


pietru2004:
Would there be possibility to implement pointer mode that would automatically show laser/target when controller is directed on screen?

Malcolmnixon:
That should be a fairly simple change. We already have the raycast to detect collisions. We'll just need to change the visibility options to show/hide/collision for the laser and target.

PR #221, opened 2022-10-25
Added more laser options to XRToolsFunctionPointer

Malcolmnixon:
This pull request implements feature #217 with the following changes:
 - Added option to only show laser pointer on collision
 - Added option to truncate laser length to collision point

The configuration options are:
image
 - Show Laser can be one of [Hide / Show / Collide]
 - Laser Length can be one of [Full / Collide]

Truncation to the collision looks as follows:
image
image


BastiaanOlij (approved):
Looks good, don't forget to update the versions document :)

PR #222, opened 2022-10-25
Implement grip threshold

BastiaanOlij:
This PR introduces a grip threshold on our pickup function.

On controllers such as the Oculus touch our grip button is an actual button which results in our logic working pretty sweet.
On controllers such as the Index it's actually a pressure sensor. The problem is that the build in threshold makes it really hard to let go of things.
So here we are implementing our own so we have control over this threshold and use the grip axis as our input.

For controllers such as Oculus Touch controllers OpenXR already provides support so unpressed is `0.0` and pressed is `1.0` and switching from the grip button to the grip axis should be transparent.

Instead of having this as a property I'm introducing this as a project setting so we can share the same value on other scripts as well.
image
It's worth contemplating doing this for other shared settings as well.

I have also moved our `Buttons` enum to our plugin script, only the pickup function uses this but we can do further cleanup of other scripts so we don't re-define this everywhere

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,5 +1,6 @@
 # 3.1.0 (Development)
 - Improvements to our 2D in 3D viewport
+- Use value based grip input with threshold
 
 # 3.0.0
 - Included demo project with test scenes to evaluate features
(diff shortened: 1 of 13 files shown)


Malcolmnixon (approved):
Looks good. Tested on the Quest 2 and everything works fine.

PR #223, opened 2022-10-25
Improved pointer demo supporting left hand with switching

BastiaanOlij:
This is a simple little improvement to the pointer demo so both hands have a laser pointer, but the user can switch between right and left hand by pressing the trigger on the desired controller.

Nice improvement for our left handed players.

I've also improved the UI a bit, added a message around how the trigger works and changed it to using VBox and HBox containers for layout.


Malcolmnixon (approved):
Looks good. I'll rebase my pointer enhancements on top of this once it goes in.

Issue #204, opened 2022-10-13
Viewport2dto3D: Add fix to make textures clearer, make it easier for devs to enable refresh throttle


teddybear082:
1) Per discussions on Godot XR discord channel, almost always it will be a bad idea in VR for the Viewport to update always.  So would be good to have an easy to use toggle at the parent level of the Viewport2dto3D scene for whether someone really wants to update always or instead wants to set a throttled update or only once update.  

2) Per discussions on Godot XR discord channel, the texture displayed on the viewport may appear to be more blurry than it needs to be without a few extra lines of code: 

                var tex = $Viewport.get_texture()
		tex.flags = Texture.FLAG_FILTER
		material.albedo_texture = tex

Example of a start of a "fixed" Viewport2dto3d is here (though still needs the option for going back to auto update or totally only a once update): https://github.com/teddybear082/godot-3d-platformer-vr/blob/main/addons/godot-xr-tools/objects/Viewport_2D_in_3D.gd

BastiaanOlij:
Both good ideas, we need to find some time for this :)

PR #224, opened 2022-10-25
Added filter and update mode to Viewport 2D in 3D

BastiaanOlij:
This PR adds three new settings to `Viewport2Din3D`
- `Filter` when enabled will set texture filtering on the viewport texture resulting in a less pixelated look
- `Update Mode` allows you to set different update modes (see below)
- `Throttle FPS` is the target FPS for updating the UI, this is only active if the `Update Mode` is set to `Update Throttled`

The available `Update Mode`s are:
- `Update Once`, this will redraw the screen one time. You can repeatedly assign this to trigger another update.
- `Update Always`, will update the screen every frame that is rendered to the HMD
- `Update Throttled`, will lower the update rate to the target FPS

Note that the throttled mode implements a target FPS, the actual FPS can be lower depending on the FPS of the main output.

Implements #204

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,3 +1,6 @@
+# 3.1.0 (Development)
+- Improvements to our 2D in 3D viewport
+
 # 3.0.0
 - Included demo project with test scenes to evaluate features
 - Standardized class naming convention for all scripts to "XRTools<PascalCaseName>"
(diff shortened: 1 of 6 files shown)


Malcolmnixon (approved):
Looks good, and much clearer!

PR #226, opened 2022-10-25
Implement poke feature and demo

BastiaanOlij:
I've got a few more things to finish here but the core works.

This implements a feature where a rigid body with a sphere collision shape follows the location that its anchored at.
This can be used for two use cases:
- interact with our `Viewport2Din3D` where we mimic mouse input where the sphere collides with the viewport
- pushing objects in the scene around

The demo only shows off the first by implementing a simple drawing tool with a touch interface. You can interact with this with your index finger, but the poke feature is not limited to fingers, it can be attached to the palm or to a pencil object.
image

I've also added some cubes from our pickup demo in this scene, you can push them around with your finger.

Do note that while you can interact with multiple poke objects (one on each hand or even each finger) the limits of the mouse emulation will confuse the Viewport2Din3D implementation, not much we can do there.

Also note that you can't really combine this together with the laser pointer as they both emulate mouse input and could fight with the poke input.


Malcolmnixon (approved):
It works well in 3.4.5 now.

PR #227, opened 2022-10-26
Fixed tool mode for Viewport2Din3D

### LINKED ISSUE #225: Viewport doesn't react to size changes in the editor
While we want to limit how much works in the Viewport2Din3D node, we do want it to react to size changes so we can see where it is positioned.

This broke when we added:
```
	# Do not initialize if in the editor
	if Engine.editor_hint:
		return
```
to `_ready`.

This should be remove and instead the `_update_*` functions should be modified to exist on editor hint if we don't want them to update.

BastiaanOlij:
This PR fixes our preview/tool mode for `Viewport2Din3D`. We now get our viewport properly sized and attached scene pre-rendered:

image

Fixes #225

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,5 +1,6 @@
 # 3.1.0 (Development)
 - Improvements to our 2D in 3D viewport
+- Fixed editor preview system for our 2D in 3D viewport
 - Use value based grip input with threshold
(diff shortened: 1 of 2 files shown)


Malcolmnixon (approved):
Good fix, works well in 3.4.5 and restores the broken viewports!

PR #229, opened 2022-10-26
Improve snap turning logic

BastiaanOlij:
## Add defaults in Project Settings
Added a snap deadzone and default snap turning settings to our project settings:

image

## Add user settings system to save user preference

Added a user settings class that saves user settings to `user://xtools_user_settings.json`

There is also a default UI class that allows developers to just embed the scene into something of their choosing:
image

In our demo application we nicely integrate this UI on a pedestal:
image

But similarly we could embed access to this UI in game. We may add a demo for this where the UI can be access on the players wrist or something like that. That's for another day though.

## By default have turn movement adhere to user/system settings

Our `MovementTurn` class now has a new turn mode setting that replaces the smooth tickbox:
image

This can be:
- Default, we will check the user settings for this (which defaults to our project settings)
- Snap, which sets this to snap turning regardless of our user setting
- Smooth, which sets this to smooth turning regardless of our user setting

## Add user setting for player height adjustment

I also added a height adjust user setting:

image

This simply increases/decreases the player height calculated in the `PlayerBody` node. This allows you to play games designed for standing up while being seated, or if you're a small/tall person adjust your in game size to make the experience more comfortable. 

*edit* I also added a project setting "player standard height". This setting allows the developer to set the height of the player they are designing the game for.
There is now a button on the user settings UI that takes the current height of the player based on the camera position, compares it to the "standard height" and calculates an adjustment value that makes the player the "standard" height.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -5,6 +5,7 @@
 - Improved pointer demo supporting left hand with switching
 - Enhanced pointer laser visibility options for colliding with targets.
 - Implement poke feature (finger interaction)
+- Improvements to snap turning
 
 # 3.0.0
 - Included demo project with test scenes to evaluate features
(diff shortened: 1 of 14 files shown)


Malcolmnixon (approved):
This works very nicely. I'm quite happy with the ability to extend the project settings window.

PR #230, opened 2022-10-26
Move our staging logic into the plugin so it becomes reusable

BastiaanOlij:
This PR moves most of the staging logic into the plugin. The scene in the demo that is loaded is inherited from the plugin scene.

All that is needed is:
- setting the start scene
- adding the ARVROrigin and ARVRCamera nodes, as the setup of this is dependent on the VR interface used, this is not in the base class.


Malcolmnixon (approved):
This plays very nicely now.

PR #232, opened 2022-10-27
Fix runtime issue due to EditorPlugin dependency

BastiaanOlij:
Moved global plugin state into a separate script as EditorPlugin is not available in runtime


Malcolmnixon (approved):
Change looks good, and works on the Quest 2

Issue #210, opened 2022-10-14
movement_climb should be one-handed


goatchurchprime:
The two handed average offset is something I've never seen in any game, partly because it will always jerk when the second hand is applied or released.  Normally the second hand takes control from the first hand.  

Also it would be feel more positive if the hand/arm had to be within the surface rather than an invisible distance from away.

Malcolmnixon:
We could change it to use the most recent grab possibly. The averaging solves some corner case - for example with single-handed grabbing what happens when:
 1. The player grabs a wall with the left hand.
 2. The player grabs the wall with the right hand and pulls themselves up a foot.
 3. The player let's go with their right hand, but is still holding with the left.

BastiaanOlij:
It's also a matter that we don't anchor the hands, they stay attached to the controller. 

Many climbing games do have two handed functionality but the moment a hand grabs something, it is no longer attached to the controller, it is now attached to the place that is grabbed, and is let go automatically when the hand moves too far away.

That said, @goatchurchprime example is also true, where the last hand that grabs determined the movement. Let go with that hand and the first hand takes over again.  Still even in that situation the hands once grabbing the wall get anchored to that place on the wall instead of following controller movement.

Malcolmnixon:
I'm reminded of a Population-one bug where a player could hold on to a wall, then run half way across the map, then I believe if they released the grip they would teleport back to the wall.

We almost certainly have a similar bug in our current implementation. If a player grabs a stationary wall with the left hand, and a climbable moving platform with their right, then they would almost certainly be carried along at the average/center of those two positions.

One-handed with "snap the grab" if too far would solve that, although as you say we want to solve the hand decoupling at the same time.

BastiaanOlij:
I still need to play Population One...

Malcolmnixon:
@goatchurchprime could you confirm this climbing behavior works as desired, and if so close the issue.

PR #235, opened 2022-10-29
Changed climbing to one-handed dominance

Malcolmnixon:
This pull request implements feature #210 by making climbing use a single dominant hand.

### DIFF
--- a/addons/godot-xr-tools/functions/movement_climb.gd
+++ b/addons/godot-xr-tools/functions/movement_climb.gd
@@ -2,21 +2,22 @@ tool
 class_name XRToolsMovementClimb
 extends XRToolsMovementProvider
 
+
+## XR Tools Movement Provider for Climbing
 ##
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):
Brilliant, I'd have to play with it a bit in game, I'm wondering how well things work if you grab with your left hand, now grab with your right, pull yourself up with your right and then let go with your right.

The dominant hand now changes back to your left hand, but it is now no-where near where you originally grabbed the wall. It may look weird if you continue to climb with your left hand when its new starting location is somewhere floating in space... 

Or do we ungrab the left hand in this case when it moves off far enough?

Malcolmnixon:
If the secondary hand moves more than 1 meter from its grab-point (either because the primary hand has moved the player, or the player moves their secondary hand) then the secondary hand grab is released. If the player then releases their primary hand then the player will fall.

Issue #153, opened 2022-06-26
Function_pickup missing an enabled property


Malcolmnixon:
Function_pickup should support enabling/disabling dynamically.

PR #236, opened 2022-10-29
Added enabled property to XRToolsFunctionPickup.

Malcolmnixon:
This pull request implements feature #153 by adding an enabled property to the XRToolsFunctionPickup class.


BastiaanOlij (approved):
Very surprised we didn't have this from the start :)

PR #238, opened 2022-10-31
Added configurable grab-points

Malcolmnixon:
Moved standard hand offset [+/- 0.03, -0.05, 0.15] from all the scenes into the hand assets.
Added grab points to the interactable demo objects.
Implemented grab-points in XRToolsPickable.
Cleanup of XRToolsPickable script and scene to remove obsolete parameters.
Modified grab cube to use precise grab.


Malcolmnixon:
The hand grab-points demonstrated in this pull request support showing the hand mesh in the editor:
image

Once hand-poses via animations is implemented, the hand-mesh should probably show the grabbed pose.

DigitalN8m4r3:
Tester this a couple of times and it is pretty robust, so far i did not encounter any issues

BastiaanOlij:
I have to find some time to play with this but dang mate, this looks like a really cool improvement!

pietru2004:
Question.
Does it allow to setup more than 2 points - example, setting 2 grab points in one position but rotated differently for each hand

Malcolmnixon:
Yes - the left and right grab points can be configured with the rotations/translations appropriate for each hand. The hammer pickable object demonstrates this if you set both hands to show in the editor-preview.

Malcolmnixon:
So the hand tuning seems to work quite nicely now. The pickable demo has teacups demonstrating:
 - Left-hand grab point
 - Right-hand grab point
 - Snap-point to saucer
 - Snap-point to stand

The basic teacup has all the grab-points invisible:
image

Enabling visibility on the left-hand grab point shows the preview hand in the grabbed position by default, although it can be toggled between open and close:
image

The same thing works on the right hand:
image

The snap-zone grab-points can also be shown, and are extended from Position3D so they show a spatial alignment (green up):
image

The hand grab-points now correctly override the associated hand with whatever XRToolsHandPoseSettings is configured.

addons/godot-xr-tools/objects/grab_points/grab_point_hand.gd
```diff
@@ -0,0 +1,131 @@
...
+	if not is_instance_valid(_grabber):
+		return false
+
+	# Ensure the pickup is a function pickup for a controller
+	var pickup := _grabber as XRToolsFunctionPickup
+	if not pickup:
+		return false
+
+	# Get the parent controller
+	var controller := _grabber.get_parent() as ARVRController
```
BastiaanOlij, line 70: I think we should add a `get_controller` function to `XRToolsFunctionPickup` to future proof this if we ever decide to make it possible to add the pickup function to another node.

BastiaanOlij (approved):
Amazing work!

addons/godot-xr-tools/objects/grab_points/grab_point_hand.gd
Malcolmnixon: Added the get_controller() to XRToolsFunctionPickup, and added its use in XRToolsGrabPointHand.

PR #239, opened 2022-11-02
Image folder cleanup and added crouching and direct movement documentation.

Malcolmnixon:
Moved climbing documentation images under /assets/img/climbing folder to keep the structure clean.
Expanded the documentation for crouching and direct movement.


BastiaanOlij:
Is it a limitation that the images for documentation need to be in the assets folder?
Or could they be moved to `_docs/img` ?
It makes sense to me to only have images used in the demo under `assets` but that is just nitpicking on my part.

BastiaanOlij:
Ah never mind! now the penny drops, so the `docs` branch contains just the documentation. Cool :)

BastiaanOlij (approved):
Huge step forward for making the documentation more maintainable.

PR #243, opened 2022-11-04
Added alternative hand models

DigitalN8m4r3:
added highpoly and lowpoly version of hand models
added skin materials
added 35 Hand Poses

Note:
The models are based on the makehuman highpoly model, it was exported and reworked further inside blender 3d, there it received a new rig and the textures where made using a base texture from makehuman and edited inside Substance Painter

Am releasing this work as Public Domain CC0


DigitalN8m4r3:
Updated with right hand shading fixed and animations seperated into .anim files

added a more suitable default pose, previous default pause was renamed to spread as it resembles spreading fingers, this means the hand poses count is now 36

fixed right hand normals shading, changed transforms and set scale inside blender to -1 on xyz
changes to the previous commit, seperated the model and animations:
saved out .anim files for left and right hand
exported left/right hand without animations to ease to reuseability

SideNote:
The .anim files are only useable in Godot 3.5
regarding this PR and Godot 4, i already saved out a pose library for each hand and once this PR gets merged, i will submit a new pr with the poselibrarys for Godot 4 and make new prefab scenes for the hands

Malcolmnixon:
I swapped the old left/right physics hands with the new left/right physics_hand_low files and it crashes. I believe it's because those hands use the "hand.gd" file rather than "physics_hand.gd". If I swap them over then the game doesn't crash.

When I then play, the scene the hands don't actually respond to the grip/trigger inputs.

Malcolmnixon:
It looks like your new hands need a new animation tree - the blend tree bone names are different.
   "index_proximal" (old) -> "Index.Proximal" (new)

I believe @BastiaanOlij may have mentioned something about bone names and possibly going to the new naming convention.

DigitalN8m4r3:
Thx, am gonna fix that... 
since the handpose pr by @BastiaanOlij
is not yet merged, will have to redo/rework.
Question would be if it is better to just close this pr and resubmit once @BastiaanOlij handpose pr is in and your multi grab points pr?

Malcolmnixon:
I think it's probably best to try and get yours in first. The added benefit of the animation library will then help with the other two pull requests.

I figure the correct approach to the blend tree is to introduce a new blend tree for the new names I think the old blend tree names come from the old godot naming convention (also available with "use_legacy_names"). At some point we may re-import the original hand with the new naming convention and switch it over to the new blend tree, or may just retire it in preference for your new hand model.

BastiaanOlij:
@Malcolmnixon the problem I ran into with legacy names is that if  you set it to false, it indeed uses the bone names as they are, but it will start messing up when you edit your hand model and export new poses.

The feeling I get, but this is just a theory, it's changing the order of the bones it is exporting and it's not matching them up by name. I could be totally wrong though, the issue could be with Blender, but I found when we stick to legacy names, it does some extra magic that makes it all a lot more stable.

My gut feeling says that its better to not use legacy names, I'm sure support for it will disappear at some point, but I'd like to find out why it's misbehaving :)

DigitalN8m4r3:
@Malcolmnixon @BastiaanOlij 
fixed remaining issues and added new blendtree with the correct uppercase namings, the alternative hands are added to the interactable/pickable/climbable scenes as discussed in via discord with @Malcolmnixon

regarding the default pose, i choose a slight grip pose for the default as it looks somewhat more natural compared to the actual default pose

Malcolmnixon:
This looks very nice, and I dropped it into by Dark City demo game just to check it out and it also worked very nicely.

For that game I was going to add some fingerless gloves (just create an alternate texture with some black painted on) but I noticed you don't have the ".blend" file checked in - although you do have the backup ".blend1" in there.

Once this is in I think we're going to have to expose an "Override Material" option in the hand.gd script so we can easily swap the hand materials without having to resort to editable children.

DigitalN8m4r3:
ye, the .blend1 is from the original hands and not from the alternates
need to clean up the blend and will submit it then, figured it doesnt rly matter right now having it inside the xr tools.

regarding fingerless gloves etc.
the idea was actualy to release a addon with multiple different types of hand meshes, one for example being fingerless gloves.

Now you might want to watch my wording here, meshes as in different models...
The way this will be easy possible and useable is because of the modular nature of the animation library.
None of the hands in this commit contain any animations, the animations where saved out using a model that had all animations in it.

This means evryone can rig their handmesh to the skeletal armature inside blender or whatever modelling application is used and export the mesh and presto.

Or wait for the addon that will contain different types of handmeshes.

BastiaanOlij:
Ok, just nitpicking on a few things.

I think instead of having your hands in `res://addons/godot-xr-tools/assets/hands` let's just move it into `res://addons/godot-xr-tools/hands`, it contains so many parts, it deserves its own dedicated folder.

We also need to determine whether these replace the current hands, or live alongside them. I'm gravitating to replacing as these are really much nicer. I'm glad for the low poly versions but am wondering if we should default with the high poly ones. That may be an issue for Quest users. 

We did this already for our build in hands, but instead of embedding the offset on the root node, make sure the root node (say `LeftHand`) always is at `(0,0,0)` and instead move the node underneath to `(-/+0.03, -0.05, 0.15)` (say `hand_l`).
This ensures that the developer doesn't have to worry about resetting the position when accidentally changed and we can later on add logic if we need to change the positioning.

Other than that, I think this is amazing work. I do agree we should merge this and then I can update my PR and finish the cleanup and make sure all the anim stuff works properly.

Don't forget to add yourself to the `CONTRIBUTORS.md` file.

BastiaanOlij:
Also you might want to put a `LICENSE.md` file in the hands folder to indicate the textures and 3D files are CC0.

DigitalN8m4r3:
Alright, gonna do the suggested and rebase with Master and move the contents to its own folder and add a license file to it along with the source blend, in a day i should be done

DigitalN8m4r3:
alright @BastiaanOlij this should be it, just commited the final update that resets transforms on all prefabs and sets the transforms on the childnode containing the handmesh and added glove materials as wished to demonstrate the material override feature that @Malcolmnixon added

Malcolmnixon (approved):
Looks good and runs nicely on my quest!

Issue #245, opened 2022-11-06
Teleport movement should not need the camera node


Malcolmnixon:
Currently XRToolsFunctionTeleport needs the camera node set. This should be replaced with a call to ARVRHelpers.get_arvr_camera()

PR #246, opened 2022-11-06
Modified XRToolsFunctionTeleport to get the origin and camera from ARVRHelpers

Malcolmnixon:
This pull request implements #245 by:
- Replace getting the origin using get_node("../..") with ARVRHelpers.get_arvr_origin(self)
- Replace getting the camera using NodePath with ARVRHelpers.get_arvr_camera(self)

### DIFF
--- a/addons/godot-xr-tools/functions/function_teleport.gd
+++ b/addons/godot-xr-tools/functions/function_teleport.gd
@@ -4,18 +4,15 @@ extends KinematicBody
 # should really change this to Spatial once #17401 is resolved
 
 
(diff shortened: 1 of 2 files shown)


BastiaanOlij (approved):
Interesting that we did this for all the movement providers but teleport, better late then never :)

Malcolmnixon:
Once teleport logic is moved into XRToolsPlayerBody, the XRToolsFunctionTeleport may well become an XRToolsMovementTeleport rather than being the one movement feature not based around movement providers!

Issue #247, opened 2022-11-06
Hand nodes should support setting the material


Malcolmnixon:
Currently setting the material on a hand requires enabling "Editable Children" and navigating the hand scene to find the MeshInstance. Instead hand material override should be an exported property of the hand.

PR #248, opened 2022-11-06
Added hand_material_override export property to XRToolsHand script.

Malcolmnixon:
This pull request implements #247 by:
- Adding a "Hand Material Override" property to XRToolsHand
- Demonstrating with a ghost-hand material added to the climbing/gliding demo scene

image
image


BastiaanOlij (approved):
Yeah that makes a lot of sense to add

Issue #191, opened 2022-09-26
Rumble Should be Added to Standard Functions


lunarcloud:
The idea is that an enable-able, configurable-intense rumble should be added to things across the tools.

Examples:
- hand is in range of pickable
- hand collides
- flying (to simulate jetpack or the like)
- grapple shoot/attach
- on landing a fall
- on jumping

I intend to work on this now that #40 / #190 are merged

BastiaanOlij:
Owh these would be nice additions for sure!

lunarcloud:
Note to self, in pseudocode:

```gdscript
dictionary<event_name: string, { magnitude: float, hands: Hands, remaining_secs: float, pausable: bool }>

set(event_name: string, magnitude: float, hand := Hands.Both, duration := -1.0, pausable := true) -> void
    // todo add to the dictionary

_process(delta: float) -> void:
    // todo rumble value for each hand is the highest magnitude for that hand
    // if paused, filter out (pausable == false)
    // if nothing, reset rumble for the hand to 0
```

lunarcloud:
Maybe provide a helper method for folks, like myself, that are writing a game that does both xr and normal - to convert the magnitude from the standard input system rumble's weak/strong magnitudes into a single one.

```gdscript

combine_magnitudes(weak: float, strong: float) -> float:
    if strong > 0.001:
        return 0.5 + clamp(strong/2, 0.0, 0.5)
    else:
        return clamp(weak/2, 0.0, 0.5)

```

lunarcloud:
Starting work over at [lunarcloud:feature/191_rumble ](https://github.com/lunarcloud/godot-xr-tools/tree/feature/191_rumble )

lunarcloud:
Seems like the Godot4 rumble may be https://docs.godotengine.org/en/latest/classes/class_xrnode3d.html?highlight=XRNode3D#class-xrnode3d-method-trigger-haptic-pulse but the docs on this API is not great

lunarcloud:
Discussed with @Malcolmnixon , we should wait for Godot XR Tools 4.3.x development to start, as using the new static variable syntax will simplify the implementation greatly for rumble-generating nodes finding the rumble manager.

lunarcloud:
working on the code in a godot 4 branch. Demo progressing...

https://github.com/lunarcloud/godot-xr-tools/tree/feature/191_godot4_rumble

[...]

PR #250, opened 2022-11-07
added rumble manager and demo

lunarcloud:
Implement #191


lunarcloud:
Feel free to be brutal. I'd rather have it right.

Also, I did not implement rumbling across the functions of the toolkit (didn't know whether that should be in here or later) just in the demo.

lunarcloud:
@Malcolmnixon wants me to create a sort of "Rumble Player" to that can find the rumble manager, and provide a means to add rumble to any function/gun/etc

lunarcloud:
After more discussion with @Malcolmnixon I will be making the rumble manager a singleton which "XRToolRumbleFeature"s added to the hands will ensure exist. This Singleton will be easier to access from scripts that want to make rumbling happen.

BastiaanOlij:
I had forgotten how primitive haptics are implemented in Godot 3 :)

At first glance I think I get the idea behind this and it seems a good approach, I'll have to find some more time to actually experiment with it.

This is the kind of enhancement that could use more detail in the OP to explain what is being added and how its supposed to be used from a user point of view (i.e. someone making a game).

BastiaanOlij:
Quick discussion in XR meeting about this as we'd like to move forward with this PR. Need to get input from more XR tool users what they are expecting out of rumble and whether this is a solution they agree with.

Would love everyones feedback.

@lunarcloud would you mind rebasing this so that the merge issues are resolved so people can test?

lunarcloud:
Superceded by #557

PR #252, opened 2022-11-09
Modified climbing demo to use the new hand

Malcolmnixon:
This pull request changes the climbing demo to use the new hand model. This allows demonstrating the hand material override better by having the alternate material put gloves on the hand as well as providing the ghost-hand effect.

image


BastiaanOlij (approved):
Cool! I'm contemplating replacing all hands in all demos with the new ones and retiring the old hands.

I do think we should use the low poly ones by default so the experience deploying to lower end hardware remains good.

PR #254, opened 2022-11-10
Update 4.0-dev to have most of the master-branch features

Malcolmnixon:
This brings 4.0-dev up to date with master as of Nov 6, 2022. It has the following items missing:
 - New hand models could not be easily ported
 - Staging detection of headset removal
 - XRCamera3D switching does not work in Godot 4.0 beta 4


Malcolmnixon:
This is a nightmare of a pull request, but if possible I'd like you to run this and make sure you don't see any glaring issues.

addons/godot-xr-tools/staging/staging.gd
```diff
@@ -0,0 +1,237 @@
...
+	await _tween.finished	
+
+	# Hide our loading screen
+	$LoadingScreen.follow_camera = false
+	$LoadingScreen.visible = false
+
+	# Turn off internal process on our FPController, the internal process
+	# of our XROrigin3D will submit its positioning data to the XRServer.
+	# With two XROrigin3D nodes we'll get competing data.
+	xr_origin.set_process_internal(false)
```
BastiaanOlij, line 179: So this is no longer needed, instead we need to make the Origin node loaded in current. We can put that bit of code in `scene_base.gd/scene_loaded()`

BastiaanOlij: That said, there is a bug in the current beta (already fixed in master) where we're not updating the starting position. In the new situation we're only informing the XRServer when the current origin node's position changes, instead of doing this every frame.

BastiaanOlij (approved):
I'll find some time over the weekend to do some proper testing but I'm happy for this to get merged as is.

PR #255, opened 2022-11-10
added alternative hands to 4.0

DigitalN8m4r3:
this PR adds the new hand models to the 4.0 dev branch

added highpoly and lowpoly version of hand models
added skin materials
added 35 Hand Poses

The hands use Animation Librarys, each hand got its own along with blendtrees


Malcolmnixon:
The fingers are having a little difficulty with the pinky distorting when fully closed - it actually invert and dents in:
image

I Just checked and it also happens on the Godot 3 versions as well. It just seems like it's more obvious due to the lighting differences, and I can't find any differences in the lighting settings. The weird edge-shadows are the problem.

Malcolmnixon:
Also could you get ghost-hands working in the climbing demo. This involves:
 1. edit /scenes/climbing_demo/materials/ghost-hand.tres
 2. Expand out the material in "Next Pass"
 3. In the Next Pass material set the albedo to /addons/godot-xr-tools/hands/textures/glove_caucasian_green_camo.png
 4. Save and apply the material to both hands Hand Material Override

DigitalN8m4r3:
aye, heading to work.. gonna do that later in 15h

and over the weekend or so am gonna do the suggested changes by @BastiaanOlij 
(the renaming of all bones etc. will mean a complete reexport of the hands and animations for both 3 & 4)

DigitalN8m4r3:
@BastiaanOlij 
via discord you mentioned that u want me to rename the bones and add additional for Potential handtracking as well. Now i want to clarify if you are aware about the bone orientation issue or requirements for handtracking?

Because Julian mentioned that the bones need to be oriented the same way as the ones from the valve handmodels. 

If i do this, it will make it almost impossible for the evryday Joe"dev" to rigg new hand Models with that kind of skeleton and it will make the current poses unuseable as a new bone orientation would mean that new poses would need to be made.

BastiaanOlij:
> @BastiaanOlij via discord you mentioned that u want me to rename the bones and add additional for Potential handtracking as well. Now i want to clarify if you are aware about the bone orientation issue or requirements for handtracking?
> 
> Because Julian mentioned that the bones need to be oriented the same way as the ones from the valve handmodels.
> 
> If i do this, it will make it almost impossible for the evryday Joe"dev" to rigg new hand Models with that kind of skeleton and it will make the current poses unuseable as a new bone orientation would mean that new poses would need to be made.

Yeah OpenXR is a bit weird here as it doesn't have a rest pose so we try and manage the delta movement ourselves. If it doesn't work by just renaming the bones, I wouldn't worry about this right now. This is something to worry about for the future ifit even can be done.

DigitalN8m4r3:
preparing commit that replaces the lowpoly mesh with a fixed version since the previous one had issues as it had too few vertices
and at the same time am restructuring, placing all prefab scenes in corresponding folders for high/lowpoly to distinguish further

Malcolmnixon (approved):
Tested on Quest/Link and the hands look very good!

PR #258, opened 2022-11-12
Create sprinting movement provider

teddybear082:
-Create XRToolsMovementSprint class

-Node should be child of FP Controller rather than any particular controller

-Allows user to select which controller to use to enable sprinting

-Allows user to select between toggle and hold modes for sprinting

-When enabled, sprinting will set both controllers' direct movement nodes' max speed variables to the sprint multiplier, which is a float clamped between 1-3x, and set by default at 2

-When sprinting is deactivated, the direct movement max speeds variables are returned to their original states (this means a player cannot be "sprinting" with one hand and "walking" with the other, without needing the information stored in the player_body node)

-Emits signals for when the player starts or stops sprinting so that, for instance, sound effects could be varied based on whether user is sprinting


teddybear082:
closing because now my editor is crashing trying to make further changes not sure why but going to redownload from master and go from there.

PR #262, opened 2022-11-14
updated alternative hands

DigitalN8m4r3:
this updates the new hand models with the following:

renamed bones to match the OpenXR standard
reworked the lowpoly mesh, since the previous version had too few vertices around the back of the hand added glove materials and models without nails

* updated with lowpoly mesh fix, restructured prefab scenes into folders

restructured prefab scenes into corresponding folders for high/lowpoly
added additional highpoly glove mesh
fixed ghost material in the climbing scene

Note:
all hand scenes are now loaded as editable children, this way future updates to the mesh wont result in a total rework of the hand scenes


DigitalN8m4r3:
it might look odd that alot got replaced, the reason is simple
as the bones where renamed and additional ones came in as well, the whole tree of saved out anims was useless as it used the old names

regarding the split of the prefab scenes, as it got crowded, it seemed the most logical approach to move the hand prefab scenes into their corresponding folders for high/lowpoly

DigitalN8m4r3:
as mentioned on discord, am just wondering what to do with the blendtree, i would save em out for each hand and replace the current animationtree, same i did in the 4.0dev version of the hands

Since the cleanup_hands pr changed some naming, i didnt wanted to touch the animationtrees as they seemed broken but functional. However if i get greenlight to change the animationtrees, gonna do it

Malcolmnixon (approved):
This works well in all scenes and the new layout matches the 4.0-dev which will make porting the new features easier.

PR #264, opened 2022-11-14
Add sprinting node

teddybear082:
-Create XRToolsMovementSprint class

-Node should be child of FP Controller rather than any particular controller

-Allows user to select which controller to use to enable sprinting

-Allows user to select between toggle and hold modes for sprinting

-When enabled, sprinting will set both controllers' direct movement nodes' max speed variables to the sprint multiplier, which is a float clamped between 1-4x, and set by default at 2.0

-When sprinting is deactivated, the direct movement max speeds variables are returned to their original states (this means a player cannot be "sprinting" with one hand and "walking" with the other, without needing the information stored in the player_body node)

-Emits signals for when the player starts or stops sprinting so that, for instance, sound effects could be varied based on whether user is sprinting

### DIFF
--- a/addons/godot-xr-tools/functions/movement_sprint.gd
+++ b/addons/godot-xr-tools/functions/movement_sprint.gd
@@ -17,7 +17,6 @@ signal sprinting_started()
 ## Signal emitted when sprinting finishes
 signal sprinting_finished()
 
-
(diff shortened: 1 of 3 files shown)


Malcolmnixon (approved):
The code looks good, and I tested it by adding the sprinting to the basic movement demo with both the hold-to-sprint and toggle-sprinting as well as left and right controller.

I do think it deserves a nice demo by itself at some point.

PR #266, opened 2022-11-15
removes obsolete .blend files and adds seperate blendtrees - currentl…

DigitalN8m4r3:
…y broken

removed obsolete .blend files since these used the old bone names fixed main menu scene, right hand was set to use nails version added seperate hand_blend_trees for each hand, previously the righthands animationnodetree had for whatever reason old bone names in it as well

Note:
since the merge of the cleanup hands, the animationtree is not visualy loading up a saved out blendtree, however this is not true since it loads it up but does not show its name under the Tree Root since the name is hardcoded in the hand.gd script but if you look close at one of the hand prefab scenes, you will see for example the following ext_resource path="res://addons/godot-xr-tools/hands/animations/left/hand_blend_tree.tres" type="AnimationNodeBlendTree" id=7]

and by uncommenting the tool part of the hand/physics_hand scripts and then loading a hand scene, going to its animationtree, the previous set blendtrees name is visible, therefore the animationtree for the hands is working but the shown name is hardcoded and even by hovering over it, it is not showing if it is a ext resource or not


Malcolmnixon (approved):
This works well.

Issue #268, opened 2022-11-16
Implement hand-pose area to override hand poses


Malcolmnixon:
With the new hand-poses support, a new XRToolsHandPoseArea should be added to support setting the poses when the hands enter an area.

DigitalN8m4r3:
Indeed this is goin to be useful, especialy considering stuff such as door handles etc. 

it might be a good idea to add a check to exclude a finger - thumb/Index..etc. only

as this might allow to perform some stuff on button press on the desired finger

PR #269, opened 2022-11-16
Added hand-pose area support

Malcolmnixon:
This pull request implements #268 by performing the following changes:
 - Added hand-pose override stack to XRToolsHand
 - Modified XRToolsHand to be an Area (the hand bounding area)
 - Modified all hand scenes to configure hand Areas
 - Added XRToolsHandPoseArea to support setting poses in the hand-pose override stack

The change to making XRToolsHand an area caused a name-collision in XRToolsPhysicsHand where there was a "collision_layer" property to set the collisions for all bones, but the area also introduced its collision_layer. As such the existing property was renamed to bone_collision_area.

Additionally making XRToolsHand an area made XRToolsInteractableAreaButton trigger on the area instead of just on the physics bones. As such XRToolsInteractableArea now has detect_areas and detect_bodies exported bool properties, and the detect_areas is off by default.

In order to demonstrate hand poses, all the demo interactable objects placed in the interactable_demo scene now have poses.

### DIFF
--- a/addons/godot-xr-tools/hands/hand.gd
+++ b/addons/godot-xr-tools/hands/hand.gd
@@ -1,92 +1,31 @@
 tool
 class_name XRToolsHand, "res://addons/godot-xr-tools/editor/icons/hand.svg"
-extends Spatial
+extends Area
 
 
+## XR Tools Hand Script
 ##
-## XR Hand Script
+## This script manages a godot-xr-tools hand. It animates the hand blending
+## grip and trigger animations based on controller input.
 ##
-## @desc:
-##     This script manages a godot-xr-tools hand. It animates the hand blending
-##     grip and trigger animations based on controller input.
-##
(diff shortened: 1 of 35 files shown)


DigitalN8m4r3:
copy pasting my comment from the issue.. 
Indeed this is goin to be useful, especialy considering stuff such as door handles etc.

it might be a good idea to add a check to exclude a finger - thumb/Index..etc. only

as this might allow to perform some stuff on button press on the desired finger

BastiaanOlij:
@Malcolmnixon were you still thinking of changing this to have the area be a function object you add to the hand to make it more in line with other features? Or is this indeed ready for review?

Seeing how key this is to the hands functioning I'm not against the current approach but it did sound last night you were on to a better solution :)

Malcolmnixon:
Cancelling this pull request as a design review by @BastiaanOlij and @lunarcloud found a better solution that doesn't force design-decisions on the hands.

PR #273, opened 2022-11-17
Added hand-pose area support

Malcolmnixon:
This pull request implements feature request #268 by adding XRToolsFunctionPoseArea and XRToolsHandPoseArea. The function sits under an ARVRController and detects when the hand enters and leaves an XRToolsHandPoseArea. Upon entry the associated poses are set for the hand, and upon exit the poses are removed.

This also fixes a "null" world environment in the main_menu scene.

The following shows a hand near the button getting the Sign Point pose:
image


Malcolmnixon:
The recent update implements pose priorities - to eventually allow grabbed items to override hand-pose areas.

addons/godot-xr-tools/functions/function_pose_area.gd
```diff
@@ -0,0 +1,108 @@
+class_name XRToolsFunctionPoseArea, "res://addons/godot-xr-tools/editor/icons/hand.svg"
```
BastiaanOlij, line 1: I'm wondering about this name, I know we have an area within but it doesn't really communicate what the class is for. What about `XRToolsFunctionPoseDetector` ?

BastiaanOlij (approved):
This is amazing stuff!

addons/godot-xr-tools/functions/function_pose_area.gd
Malcolmnixon: Renamed to XRToolsFunctionPoseDetector

PR #276, opened 2022-11-18
Function pickup modified to allow child-ing to other scene nodes

teddybear082:
-Update function pickup to allow selection of controller node from anywhere in scene tree, allowing function pickup to be child of a node other than the controller

-This may be useful for using the XR Avatar project, other full body projects, "physics body" projects or projects where the meshes for interactions are more obscure such as a claw object, such that objects being "held" by the controllers directly rather than as a child of the mesh may create a "floaty" effect to the object pickup.  In other words, where the pickup point may meaningfully diverge from the controller location, being able to set the function pickup to the intended location is helpful, such as to a hand bone attachment on a body mesh or the exact intended point of a moving claw.

-Because default functionality still assumes function pickup is a child of the controller, this change does not damage any pre-existing projects and is non-breaking.  Someone who does not select a controller node but has the function_pickup as a child to the controller does not have to do anything else for this to work out of the box (that is, choosing a node path is purely optional).

-There is an output warning statement if the dev mistakenly does not child the function pickup to a controller and also does not choose a nodepath for a controller, and in that event the function pickup is disabled.

### DIFF
--- a/addons/godot-xr-tools/functions/function_pickup.gd
+++ b/addons/godot-xr-tools/functions/function_pickup.gd
@@ -26,6 +26,9 @@ const MAX_GRAB_DISTANCE2: float = 1000000.0
 ## Pickup enabled property
 export var enabled : bool = true
 
+## Controller tied to this pickup function
+export (NodePath) var _pickup_controller_node_path
+
 ## Grip controller button
 export (XRTools.Axis) var pickup_axis_id = XRTools.Axis.VR_GRIP_AXIS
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
I'm on the fence with this one.

On the one hand the whole idea is to keep things simple and make sure objects are children based on what they operate on. Malcolm and I have done a lot of work in simplifying this. Adding this nodepath will confuse new users of the toolkit who don't understand that this is an alternative use case and are going to wonder how they should set this up. Sadly people hate reading manuals.

The flip side is that I get the use case. Due to the nature of skeletons and full body meshes this will often be a separate node tree alongside the ARVROrigin tree, which means there is a disjoint between where functions that have a visual impact are, and where the controller nodes they operate on are.

A better approach may be that instead for us checking if the parent node is an ARVRController node, we check if the parent node has an `get_is_active` and `get_joystick_axis` method and if so we allow it. 
Now you can add a bone attachment node for each hand in the avatar model and child the function to that. Just implement the two methods there and have them just call the methods on the relevant ARVRController nodes, those are I assume already known with the avatar as the IK targets will be following those.  

This way we hide the complexity of the solution from the end user, and probably open a few other doors where we have objects that follow the hands instead of being direct children

BastiaanOlij:
Note that this is similar to #326 and there is further discussion on the subject there.

PR #278, opened 2022-11-19
updated Hands with bone roll 0 to support future Handtracking work

DigitalN8m4r3:
additional stuff that this pr adds:
new glove hand models
(note: they got no textures yet, there will be a update for those)

changes:
since the animations already contain a sign pose, the sign_point was reverted

changes to the meshes:
to account for handtracking, additional vertices where added around the index/middle/ring fingers


DigitalN8m4r3:
if there should be a question regarding why all animation poses are replaced again,
the answer is simple, by resetting the bone roll to 0 all previous animations where useless and therefore i had recreate them from scratch again

Malcolmnixon (approved):
Tested in the demo scenes, and with visible collision shapes to verify the hand-bones are still where they expect to be. Everything looks good.

PR #281, opened 2022-11-20
Fixed loading and saving of user settings

Malcolmnixon:
The user settings were not being loaded or saved correctly, and would crash if the settings data was corrupt.

### DIFF
--- a/addons/godot-xr-tools/user_settings/user_settings.gd
+++ b/addons/godot-xr-tools/user_settings/user_settings.gd
@@ -20,31 +20,41 @@ func _load():
 	# First reset our values
 	reset_to_defaults()
 
-	# Now attempt to load our settings file
-	var file = File.new()
+	# Skip if no settings file found
+	var file := File.new()
 	if !file.file_exists(settings_file_name):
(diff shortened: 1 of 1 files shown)


addons/godot-xr-tools/user_settings/user_settings.gd
```diff
@@ -20,31 +20,41 @@ func _load():
...
+	# Skip if no settings file found
+	var file := File.new()
 	if !file.file_exists(settings_file_name):
 		return
 
+	# Attempt to open the settings file for reading
 	if file.open(settings_file_name, File.READ):
-		var text = file.get_as_text()
-		file.close()
+		# File open failed with an error
```
BastiaanOlij, line 30: If you successfully open the file you return?

I think this should be
```
if !file.open(settings_file_name, File.READ):
  # File open failed with an error
  return
```

BastiaanOlij (approved):
Happy for this to be merged as a cleanup (well with the introduced bug fixed), but I'm not entirely sure what use case this fixes?

Issue #263, opened 2022-11-14
New find node script in ARVRHelpers may be broken with custom XRTools classes


teddybear082:
I think the "is_class" function does not work to return custom class names.

I tested when creating my sprint movement functions, adding the code to search for direct movement providers repeatedly crashed the editor.  When moving that code to the process function so that it would at least run, the errors came from the new _find_child script. According to documentation, is_class will return true only for the base Godot class not any custom classes.  There is an "is" keyword, i.e., if node is class_name, but I'm not sure if this works either in the expected way; I couldn't test because "type" is locked to a String type variable and if is keyword works, the class name would be just a class name and not a string (I think).

Malcolmnixon:
Yah, apparently this is a common issue with gdscript. The best solution appears to be overriding is_class in all classes. Something like:

```gdscript
func is_class(name : String) -> bool:
    return name == "CustomClass" or .is_class(name)
```

teddybear082:
Also instead of overwriting is_class in every XRTools function, maybe you could utilize the Node "get_filename()" method instead.  E.g., in lines like 145 of ARVRHelpers, you could use if node.get_filename() == type instead of node.is_class(type)), and in the function that is calling this, type could be a string maybe of the absolute path to the script for the relevant XRTools function, e.g.,

ARVRHelpers.find_child(_left_controller, "*", "res://addons/godot-xr-tools/functions/movement_direct.tscn") 

I think this would allow you to keep "type" as a String, just it would have to be documented that to use the function you need to drag over the path to the XRTools functions you are searching for.  Not sure if that would work but just an idea.

Malcolmnixon:
I'm trying to model these functions on the new Godot 4 find_children() method so I'd like to keep the contract essentially identical, however there too it looks like custom classes isn't supported - at least not until https://github.com/godotengine/godot/pull/62273 goes in.

Until the standard *::is_class() recognizes class_name, the only option appears to be to manually add it.

Malcolmnixon:
According to https://github.com/godotengine/godot/issues/21789#issuecomment-1312540449 it looks like even the workaround of overriding is_class() is not supported in Godot 4. As the project is trying to simultaneously support Godot 3 and Godot 4 with minimal differences, it seems like the advanced node-search capability can't be supported yet.

I think for now it's best we stick with the user manually specifying NodePath properties to the appropriate nodes. It's a shame as this results in significantly more manual configuration steps.

[...]

PR #282, opened 2022-11-20
Fixed advanced node-find routines

Malcolmnixon:
This pull request fixes issue #263 by adding find-support to all XRTools classes. This change includes:
- Adding is_class() override to XRTools classes to support node-search routines
- Moving static node-search routines from ARVRHelpers to XRTools class
- Adding static ARVRHelpers.get_arvr_controller() method
- Adding static find methods to key XRTools classes

The XRTools classes now make use of these find routines in many areas, which results in the following changes:
- Numerous bespoke search routines could be removed simplifying code
- Many nodes can find their dependency nodes using onready assignment - simplifying their implementation
- The new ARVRHelpers.get_arvr_controller allows for more flexible node hierarchies under controllers

### DIFF
--- a/addons/godot-xr-tools/effects/vignette.gd
+++ b/addons/godot-xr-tools/effects/vignette.gd
@@ -1,4 +1,5 @@
 tool
+class_name XRToolsVignette
 extends Spatial
 
 export var radius = 1.0 setget set_radius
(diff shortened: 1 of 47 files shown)


BastiaanOlij (approved):
Shame the `is_class` workaround is needed for now but atleast this way it works for current releases of Godot.

PR #283, opened 2022-11-21
Added a sprinting demo

Malcolmnixon:
This pull request adds a sprinting demo scene with a race track as well as a few ramps and climbing walls.

Additionally XRToolsFunctionPickup was modified to be a tool script so is_class() works in the editor. This was necessary to satisfy the _get_configuration_warning() of XRToolsMovementClimb checking that left and right pickups are available.

image


BastiaanOlij (approved):
Cool!

PR #286, opened 2022-11-23
Implement solution to use stored blend trees

BastiaanOlij:
With the changes we made that allows us to update the poses and see the results in the editor, we had to make sure the blend tree was made unique. This required a "deep copy" to make it reliable.

The downside of this was that we lost the link with the blend trees resource files making it harder to make changes as each hand scene now has a unique copy.

This PR solves this by recording the required blend tree file as an export variable on the hand script, and loading and duplicating that blend tree and assigning it to the animation tree node.

As an added bonus this allows for users to provide their own blend trees if they choose to do so.

### DIFF
--- a/addons/godot-xr-tools/hands/hand.gd
+++ b/addons/godot-xr-tools/hands/hand.gd
@@ -16,6 +16,8 @@ extends Spatial
 ## Signal emitted when the hand scale changes
 signal hand_scale_changed(scale)
 
+## Blend tree to use
+export var hand_blend_tree : AnimationNodeBlendTree setget set_hand_blend_tree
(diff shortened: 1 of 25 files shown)


BastiaanOlij:
Note that if you edit the blend tree files, any open hand scenes won't reflect the change until they are re-opened.

DigitalN8m4r3:
seems to work as @BastiaanOlij intended, tested it with statemachine as well as wit the basic animationtree/handblendtree

BastiaanOlij:
With Malcolm not feeling well and Digital having checked this, I'm merging this :)

PR #287, opened 2022-11-23
update ghost_hand.material to be in xr tools hand folder

DigitalN8m4r3:
this pr removes the ghost_hand.material from scenes/climbing_gliding_demo/materials/ and sets it up to be inside the godot-xr-tools/hands/material folder

fixes previously made gloves_caucasian_dark_camo.material depend on the scenes/climbing_gliding_demo/materials/ghost_hand.tres


BastiaanOlij (approved):
Looks good to me

PR #290, opened 2022-11-29
change snap turn logic to one turn per joystick jog

goatchurchprime:
The current step_turn_delay implementation on the snap-turn (which works like a repeat-key) is unconventional and can cause discomfort.

This changes it to match the behavior found in most VR games (eg HalfLifeAlyx) where you get one snap-turn when you pull the stick all the way across to the left or right, and then you only get a second turn after you back it off at least halfway to the centre.  

There is also a change to the smooth turning so that the deadzone width is subtracted from the joystick value to enable turns as slow as you want them.  A deadzone on the stick reading is always necessary because there would otherwise be a constant drift with the joystick never being exactly at 0.


goatchurchprime:
Slight correction.  AltspaceVR does the repeat-key snap-turning, but not VR-chat

BastiaanOlij:
My experience with most games with snap turning was that it worked like we did, mind you I turn it off nowadays so maybe things have changed since I actively used it but I actually haven't run into this way of it working (I also don't play vrchat :P ).

That said, I can see the comfort benefit. 

IMHO this should be an option so the user can switch between continues snap turn and single snap turn.

Malcolmnixon:
I agree - it's a nice change, but it should be optional seeing as we don't want to break backwards compatibility for existing users. A snap-turn mode enumeration of Single or Repeating.

Malcolmnixon:
When the physics service rate is high, the current turning code becomes hard to use. When the user releases the joystick from a snap-turn - it can momentarily swing to the other side. What happens is the provider first sees it as being "smaller" than the dead-zone and so clears the delay, and then it swings the other way and the player snap-turns back.

As such we'll probably add the single-snap-turn option at the same time.

Malcolmnixon:
The current implementation can be tuned to match both styles. Its current behavior does the repeating style; but one-shot feels comfortable by:
 - Setting the project "Snap Turning Deadzone" high (above 0.5)
 - Setting the "Step Turn Delay" larger (0.5 or more)

My issue with snap-back was caused by a "Snap Turn Deadzone" that was too low.

goatchurchprime:
Changed so if you set step_turn_delay to zero then you only get the single snap version.

These option factors should probably be in the XRToolsUserSettings, but I don't see an easy way to add them

BastiaanOlij:
> Changed so if you set step_turn_delay to zero then you only get the single snap version.

Seems like a very logical solution to me

> These option factors should probably be in the XRToolsUserSettings, but I don't see an easy way to add them

I'll have a look at adding that when I have some free time, I don't see this as a show stopper.

BastiaanOlij (approved):
lgtm, don't forget to add your name to Contributors.md

It would also be good if you squash all commits.

goatchurchprime:
replaced by https://github.com/GodotVR/godot-xr-tools/pull/331

PR #292, opened 2022-12-02
add option to rotate player on arms roll movement, during glide action

surreal6:
This PR adds an option to enable yaw movement for the player on roll action with hands. You can turn by rolling left or right without moving your feets. There is also a new exported variable to multiply the turning angle.

By rolling i mean rotate the torso by moving one hand up and the other down, as in a plane or glide.

I created an intermediate variable called left_to_right_vector to avoid repeating the left_to_right vector calculation.

I copied _rotate_player function from MovementTurn node script to apply the movement. Not sure if it's the most convenient way to do this.

It's my first collaboration, so please let me know if you find it interesting or if  i need to fix something.

### DIFF
--- a/addons/godot-xr-tools/functions/movement_glide.gd
+++ b/addons/godot-xr-tools/functions/movement_glide.gd
@@ -54,6 +54,12 @@ export var horizontal_slew_rate : float = 1.0
 ## Slew rate to transition to gliding
 export var vertical_slew_rate : float = 2.0
 
+## glide rotate with roll angle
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Look good at first glance, will leave it up to @Malcolmnixon to properly review and approve as he wrote the glide logic originally.

Malcolmnixon (approved):
This works quite nicely - a natural tilt-and-turn feel!

PR #293, opened 2022-12-03
Make new XROrigin3D current after loading new scene

dsnopek:
Without this, when using "staging", the old `XROrigin3D` will remain current, and even though locomotion still "works", it'll move around the old `XROrigin3D` rather than the new one.

### DIFF
--- a/addons/godot-xr-tools/staging/scene_base.gd
+++ b/addons/godot-xr-tools/staging/scene_base.gd
@@ -48,6 +48,7 @@ func scene_loaded():
 	
 	# Make sure our camera becomes the current camera
 	$XROrigin3D/XRCamera3D.current = true
+	$XROrigin3D.current = true
 	
 	# Center our player on our origin point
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
Note, this requires https://github.com/godotengine/godot/pull/69531 to work.

I believe @Malcolmnixon already has this in his list of things he was taking along in the next conversion pass..

Malcolmnixon:
I've checked this works well with https://github.com/godotengine/godot/pull/69531. I'm wondering whether we should wait until this gets in to Godot 4.0 Beta 8, or just get it in now and have people do a manual Godot build while waiting.

BastiaanOlij:
I think we should wait on beta 8, Akien is now doing weekly builds.

Malcolmnixon:
I can confirm this PR works correctly with Godot 4.0 beta 8.

BastiaanOlij (approved):

PR #298, opened 2022-12-10
Modified the teleport title material to perform tiling in the shader

Malcolmnixon:
The teleport titles don't render correctly in Godot 4.0 when running in GLES3 mode - specifically the shader was written assuming the UV values could exceed 1.0 and the texture would wrap. Now the shader performs the fract() operation to tile internally so any texture can be used.


BastiaanOlij:
Hmmm, that feels wrong, this should be controlled by properties on the texture or on the shader. Introducing a function on the UV in the fragment shader on mobile will punish you as you can't do texture prefetching.

So this sounds like a bug in Godot 4 itself that must be reported and fixed properly.

Malcolmnixon:
Closing this, as the true fix is to switch the sampler to repeat.

PR #299, opened 2022-12-10
Modified the teleport title to perform texture-tiling internally

Malcolmnixon:
This modified the 3.x version of the teleport title shader to work the same as the 4.0 version in that the shader performs the texture-tiling rather than relying on tiling being turned on for the texture.


Malcolmnixon:
In that case we probably want to make the title images all powers of two to silence the gles2 warnings.

Malcolmnixon:
Reverting this as the issue was identified in the godot 4 shader.

PR #300, opened 2022-12-14
Modified the teleport title sampler to enable repeat.

Malcolmnixon:
This pull request enables repeat in the texture sampler as the default value did not enable repeating in gl_compatibility mode.
image

### DIFF
--- a/assets/meshes/teleport/teleport_title_material.tres
+++ b/assets/meshes/teleport/teleport_title_material.tres
@@ -18,6 +18,7 @@ description = "These nodes add the image and scrolling UVs to sample the title t
 [sub_resource type="VisualShaderNodeTexture2DParameter" id="5"]
 parameter_name = "Title"
 texture_type = 1
+texture_repeat = 1
 
 [sub_resource type="VisualShaderNodeInput" id="7"]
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):
lgtm

assets/meshes/teleport/teleport_title_material.tres
```diff
@@ -18,6 +18,7 @@ description = "These nodes add the image and scrolling UVs to sample the title t
 [sub_resource type="VisualShaderNodeTexture2DParameter" id="5"]
 parameter_name = "Title"
 texture_type = 1
+texture_repeat = 1
```
BastiaanOlij, line 21: Was this a left over from the previous try?

Malcolmnixon: No, I threw that branch and pull request away.

It looks like the Visual Editor put this attribute on the node properties for the node-view, and then also updated the embedded copy of the generated shader code.

BastiaanOlij: Well if its the visual editor that adds that in then it's fine I guess :)

PR #301, opened 2022-12-16
Add support for WebXR

dsnopek:
This adds support for WebXR in a very simplistic way.

The main differences between OpenXR and WebXR are:

- In how WebXR is initialized (it has to be done by user interaction and is asynchronous)
- With how input is handled (mainly, in WebXR, all controllers are assumed to have BOTH a thumbstick and a trackpad, even if they don't actually, and there's no way to check)

So, this PR adds:

- An `XRToolsWebXR` autoload that can be used to detect if WebXR is available, and assist with mapping the thumbstick/trackpad issue
- An addition to user settings to allow configuring whether to use the thumbstick or trackpad as primary or secondary, or an "Auto" mode where it checks which one has input (which is imperfect because the player may just not be pressing it)
- A UI scene that can be added to provide an "Enter VR" button to initialize WebXR. In a real game, the developer would probably want to make their own customized UI, but this is a default to help folks get started.

Bastiaan and I have discussed eventually generalizing the "OpenXR Action Map" system into an "XR Action Map" system to fix the input differences, but that's Godot 4.1 material at least. This is a temporary fix for now.

Please let me know what you think!


BastiaanOlij:
I'm not sure if I'm a fan of this as a feature in XR Tools itself. It needlessly complicates matters for a very singular case.

The whole reason we have the action selectable in all the functions is so you can change it to what you need it to be for your game. While the input system in Godot 3 was fixed due to it's history, in Godot 4 it's fully configurable. The default action map is just that, a default. That the default currently has a "primary" and "secondary" input action for our joystick/thumbstick, is just that, a default.

For someone who wishes to make an XR game in Godot it's actually advisable to not use the default action map at all, but create their own action map with actions that are meaningful to their game. The "correct" methodology in OpenXR then is to say create an action called "movement", use that action for the direct movement function, and then map movement to whatever input you want to use, be it a thumbpad, joystick, d-pad or whatever other inputs may be supported at some time.

This bring us to the problem that WebXR currently falls outside of this approach as there is no way to configure an action map for webXR, something we did discuss as something to improve in the future but for now we have to work on the assumption it doesn't exist. The frustrating and dumb thing about WebXR is that we don't know which inputs are actually available, even if we have a "movement" action, WebXR doesn't allow us to map joystick on controllers that support a thumbstick and thumbpad on controllers that don't, we don't seem to get this information if I understand correctly. So it is up to the user in this case to determine whether they want to use the thumbpad or the thumbstick as input.

I feel the right solution is for WebXR to have a method that sets the action used depending on user preference but does so simply by changing the action set on the functions. Not by logic inside of the functions themselves.
The smart approach here is that when a user is targeting both WebXR and OpenXR, is for them to create an OpenXR action map that also has a thumbstick and thumbpad action to mimic the same behavior webXR has.

Finally, if one day we do get to moving the action map system to the core and having WebXR use the action map system (but then with bindings parsed in Godot instead of in the XR runtime) we should make it possible to bind both thumbstick and thumbpad to the same action and have it either work additive or have some way of detecting which one is the first to return a non-zero value and pick that as leading.

dsnopek:
@BastiaanOlij Thanks for checking out the PR!

> I feel the right solution is for WebXR to have a method that sets the action used depending on user preference but does so simply by changing the action set on the functions. Not by logic inside of the functions themselves.
> The smart approach here is that when a user is targeting both WebXR and OpenXR, is for them to create an OpenXR action map that also has a thumbstick and thumbpad action to mimic the same behavior webXR has.

Would you be OK with modifying the demos in XR Tools to use this approach? That way we can model for other people how to use XR Tools in a way that supports both OpenXR and WebXR.

Also, this PR includes the UI scene that developers can optionally use to initialize WebXR. Is that something you'd be OK with including in XR Tools?

dsnopek:
@BastiaanOlij Ok, I just push some changes. How does this sound?

- Removed all changes to the functions scenes
- Removed the WebXR singleton
- The optional UI to enter WebXR remains
- The changes to the settings for configuring WebXR thumbstick vs trackpad remain, but I redid the autodetection to work as you described: which ever we detect input from first will be selected. It also adds `webxr_primary_changed` signal that the application can connect to so they can respond to the user changing the action.
- Adds a `scenes/scene_base.gd` which extends `XRToolsSceneBase`, connects to the `XRToolsUserSettings.webxr_primary_changed` signal, and reconfigures the demo scenes for whether thumbstick/trackpad was selected or detected

So, there's still some small changes in the addon itself, mostly around settings, but I don't think it should bother anyone who is using OpenXR rather than WebXR. It the extra settings are easily ignorable, and the UI bits only appear if WebXR is actually in use.

Please let me know what you think!

addons/godot-xr-tools/staging/scene_base.gd
```diff
@@ -22,40 +22,41 @@ signal load_scene
...
 func is_xr_class(name : String) -> bool:
 	return name == "XRToolsSceneBase"
 
-
 func center_player_on(p_transform : Transform3D):
 	# In order to center our player so the players feet are at the location
 	# indicated by p_transform, and having our player looking in the required
 	# direction, we must offset this transform using the cameras transform.
-	
+
```
BastiaanOlij, line 36: Totally unrelated to this PR but maybe we should add the pblank scripts from Godot to our plugin, or @Malcolmnixon would the linter take care of this if we go through the pain of cleaning everything up for it?

addons/godot-xr-tools/user_settings/user_settings.gd
```diff
@@ -22,19 +46,42 @@ func reset_to_defaults() -> void:
...
+	return webxr_primary
+
+
 ## Save the settings to file
 func save() -> void:
 	# Construct the settings dictionary
 	var data = {
 		"input" : {
-			"default_snap_turning" : snap_turning
+			"default_snap_turning" : snap_turning,
```
BastiaanOlij, line 83: I would create a root node "webxr" and place all webxr settings in there instead of in "input"

addons/godot-xr-tools/user_settings/user_settings_ui.tscn
```diff
@@ -9,125 +9,106 @@ theme_override_font_sizes/font_size = 12
...
 theme_override_font_sizes/font_size = 12
 text = "Snap turning:"
 
 [node name="SnapTurningCB" type="CheckBox" parent="Input/SnapTurning"]
-offset_left = 82.0
-offset_right = 106.0
-offset_bottom = 24.0
+layout_mode = 2
+
+[node name="WebXR" type="HBoxContainer" parent="Input"]
```
BastiaanOlij, line 25: Same for the webXR UI items, I'd move them into a webXR tab

addons/godot-xr-tools/objects/viewport_2d_in_3d.tscn
```diff
@@ -7,13 +7,13 @@
 resource_local_to_scene = true
 size = Vector2(3, 2)
 
-[sub_resource type="ViewportTexture" id="ViewportTexture_fludi"]
+[sub_resource type="ViewportTexture" id="ViewportTexture_bfkdf"]
```
BastiaanOlij, line 10: Hmm, we need to find a way to filter out these changes, I thought the whole point of these id's was that they stay unique and I don't know why Godot keeps updating them when you clone anew. I'm going to ask about this. David I know its a PITA but any chance you can review which files only have these types of changes and remove them from the commit?

BastiaanOlij:
Looks good @dsnopek , just a few nitpick things. @Malcolmnixon I think you should weigh in on this too

scenes/scene_base.gd
```diff
@@ -0,0 +1,26 @@
+extends XRToolsSceneBase
+
+func _ready():
+	super._ready()
+
+	var webxr_interface = XRServer.find_interface("WebXR")
+	if webxr_interface:
+		XRToolsUserSettings.webxr_primary_changed.connect(self._on_webxr_primary_changed)
```
BastiaanOlij, line 8: Shouldn't this call `_on_webxr_primary_changed` atleast once to setup the right action first time the scene is loaded?

BastiaanOlij: I would also change `if webxr_interface` to `if webxr_interface and webxr_interface.is_initialised()`, the webXR interface will always be there, even if it is not being used. in its present form this code would also run if OpenXR is used.

Malcolmnixon:
I'm OK with this pull request - once the conflicts are resolved ;)

I've got a PR to add gdlint checking to the master/3.x branch; and I'll be adding it to the 4.0-dev branch soon.

addons/godot-xr-tools/staging/scene_base.gd
dsnopek: Ah, yeah, sorry about this! vscode likes to remove whitespace from lines that have only whitespace

addons/godot-xr-tools/objects/viewport_2d_in_3d.tscn
dsnopek: This file does have other changes! It's this line below:

```
gui_embed_subwindows = true
```

That's necessary to allow any controls with popups to work, like the `OptionButton` that's used by the WebXR settings. I don't think there are any files in this PR that include only ID changes.

scenes/scene_base.gd
dsnopek: > Shouldn't this call _on_webxr_primary_changed atleast once to setup the right action first time the scene is loaded?

It does that! It's the next line after the ones you quoted above. :-)

> I would also change `if webxr_interface` to `if webxr_interface and webxr_interface.is_initialised()`, the webXR interface will always be there, even if it is not being used.

It's normal to enter and exit WebXR, sometimes multiple times. I think this really does need to be setup even if WebXR is just available and not initialized yet, because you want it to be ready for when WebXR is entered.

In fact, checking if WebXR is initialized here will break the demo, because when the web page is first loaded, you're not in WebXR mode, but the staging scene is already running.

>  in its present form this code would also run if OpenXR is used.

No, OpenXR and WebXR actually can't co-exist in the same build. WebXR will _only_ be compiled in a web build, and OpenXR will _never_ be compiled in a web build.

Another way to think of it, is that all the checks to see if WebXR is available are also implicitly checking if we are in a web build.

addons/godot-xr-tools/user_settings/user_settings.gd
dsnopek: Done in latest push!

addons/godot-xr-tools/user_settings/user_settings_ui.tscn
dsnopek: Done in latest push!

dsnopek:
@BastiaanOlij @Malcolmnixon The conflicts are resolved, and all review has either been addressed in the latest push, or responded to in comments. This is ready for another round of review!

scenes/scene_base.gd
BastiaanOlij: ah offcourse, I keep forgetting webXR is very specific to the html platform and only available there.

addons/godot-xr-tools/user_settings/user_settings_ui.tscn
```diff
@@ -3,127 +3,130 @@
...
 script = ExtResource("1")
 
 [node name="Input" type="VBoxContainer" parent="."]
-anchors_preset = 15
-anchor_right = 1.0
-anchor_bottom = 1.0
-offset_top = 25.0
-grow_horizontal = 2
-grow_vertical = 2
+layout_mode = 2
```
BastiaanOlij, line 14: these btw I think are Godot 3.5 features. @Malcolmnixon I think it's time to make 3.5 our minimum version?

BastiaanOlij (approved):
I'm for approving this, the id changes are annoying but not a huge barrier.

addons/godot-xr-tools/user_settings/user_settings_ui.tscn
Malcolmnixon: Agreed - this may have gone in to 4.0; but I'm trying to keep the master and 4.0-dev branches in sync.

PR #303, opened 2022-12-19
updates hands to bring 4.0 on par with 3.5 - hands

DigitalN8m4r3:
this pr updates the hands to be on par with the 3.5 hands at the same time it paths the way for the upcoming hand pose stuff

for the sake of the hand pose
the animations were saved out as .res files, as in 3.5 we used .anim files.

Now the idea here is to change the filename ending in the loading code of the hand pose stuff and this way to allow the same functionality as in 3.5


DigitalN8m4r3:
changes this to draft to test it out, since the above is more a theory

at the same time the scenes are not yet containing the exact same hand prefabs as in 3.5

Malcolmnixon:
I've checked the animations here, and they can be used to replicate the same hand-pose mechanics as are present in godot-xr-tools 3.5

PR #305, opened 2022-12-23
Added support for different player "up" orientations.

Malcolmnixon:
This pull request allows for different "up" orientations for the player. The changes are:
 - Modify XRToolsPlayerBody to apply gravity/control based on the ARVROrigin up direction
 - Modify XRToolsPlayerBody to support orientation providers which control the player orientation
 - Added XRToolsOrientationGround provider to orient to the ground under the players feet
 - Added XRToolsOrientationArea provider to orient based on orient-control areas
 - Added XRToolsOrientAreaPath to orient the player along a path
 - Added XRToolsOrientAreaPoint to orient the player to a point
 - Added XRToolsOrientAreaUniform to orient the player in a uniform vector
 - Moved common code from movement and orientation providers to XRToolsPlayerProvider
 - Modify XRToolsMovementClimb to work in the players orientation
 - Modify XRToolsMovementFlight to work in the players orientation
 - Modify XRToolsMovementGlide to work in the players orientation
 - Added Origin Gravity demo scene
 - Added Sphere World demo scene

Additionally the following cleanup was performed:
 - Modify XRToolsPlayerBody to have common rotation code (extracted from turn and glide)

image
image


BastiaanOlij:
Absolutely love the idea but as I mentioned on Discord, not a fan of the approach. We're reinventing the wheel of the physics engine here and as a result we're only have parts of our own implementation playing by our own rules while other parts of the physics engine plays with it's rules.

In order to create the behavior we're after in Godots build in physics engine you use area nodes that override the gravity within that area from the default, see [space_override](https://docs.godotengine.org/en/stable/classes/class_area.html#class-area-property-space-override).
To create spherical gravity you enabled [gravity_point](https://docs.godotengine.org/en/stable/classes/class_area.html#class-area-property-gravity-point) and set the center of gravity by setting [gravity_vec](https://docs.godotengine.org/en/stable/classes/class_area.html#class-area-property-gravity-vec)

Note that with `gravity_point` set to false `gravity_vec` works the same as the implementation offered here.

We would still need to change our functions in similar ways but we're now getting our information from the same source the physics engine is getting it from resulting in objects that use the build in logic from behaving the same as our player. Ergo if you're at the bottom of your planet and you throw something, they don't fall off into space.

Malcolmnixon:
Updated the pull request to use physics gravity rather than orientation providers. The movement providers can now access both "up_player" and "up_gravity" information as appropriate for the motion. Movement providers can also perform a pre-processing step of changing the gravity - this is necessary for the wall-walking movement provider to have the player stick to the walls, while still supporting all other actions (running, jumping, etc).

BastiaanOlij (approved):
Very cool addition, this is pretty cool

PR #310, opened 2022-12-29
Clean godot-xr-tools to have no warnings against gdlint

Malcolmnixon:
This pull request cleans up all gdlint (https://github.com/Scony/godot-gdscript-toolkit) warnings.

In some cases this took complete reordering of the contents of the files to match the code ordering from the gdscript style guide - https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#code-order

### DIFF
--- a/.github/workflows/gdlint-on-push.yml
+++ b/.github/workflows/gdlint-on-push.yml
@@ -1,0 +1,25 @@
+# Workflow to automatically lint gdscript code
+name: gdlint on push
+
+on:
+  [push, pull_request]
+
+jobs:
+  gdlint:
+    name: gdlint scripts
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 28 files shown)


BastiaanOlij (approved):
lgtm, I do think we should add in the CI changes to make sure we start checking linting on PRs.

Malcolmnixon:
Added the gdlint-on-push action to run gdlint on the addons/godot-xr-tools folder.

BastiaanOlij:
Looks like it's running happily. At some point we should also look into implementing it as a git hook so it runs automatically before committing though that will require a bit more investigation as you only want to run it on changed files.

BastiaanOlij (approved):

PR #312, opened 2023-01-04
Added initial WebXR entry UI and export target.

Malcolmnixon:
This pull request adds WebXR support to the godot-xr-tools demo project.

Note - this turns on Thread support in the WebXR HTML5 export presets as the godot-xr-tools project is set up to use resource_queue to load resources in a worker thread.

### DIFF
--- a/addons/godot-xr-tools/staging/staging.tscn
+++ b/addons/godot-xr-tools/staging/staging.tscn
@@ -1,8 +1,10 @@
-[gd_scene load_steps=8 format=2]
+[gd_scene load_steps=10 format=2]
 
+[ext_resource path="res://addons/godot-xr-tools/misc/vr_common_shader_cache.tscn" type="PackedScene" id=1]
 [ext_resource path="res://addons/godot-xr-tools/staging/staging.gd" type="Script" id=2]
 [ext_resource path="res://addons/godot-xr-tools/staging/loading_screen.tscn" type="PackedScene" id=3]
 [ext_resource path="res://addons/godot-xr-tools/staging/fade.gdshader" type="Shader" id=4]
+[ext_resource path="res://addons/godot-xr-tools/xr/start_xr.tscn" type="PackedScene" id=5]
 
 [sub_resource type="QuadMesh" id=4]
 custom_aabb = AABB( -5000, -5000, -5000, 10000, 10000, 10000 )
@@ -29,9 +31,27 @@ material/0 = SubResource( 3 )
 environment = SubResource( 2 )
 
 [node name="LoadingScreen" parent="." instance=ExtResource( 3 )]
-splash_screen = null
 progress = 0.0
 
 [node name="Scene" type="Spatial" parent="."]
 
 [node name="Tween" type="Tween" parent="."]
+
+[node name="ARVROrigin" type="ARVROrigin" parent="."]
+
+[node name="ARVRCamera" type="ARVRCamera" parent="ARVROrigin"]
+transform = Transform( 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 1.8, 0 )
+
+[node name="VRCommonShaderCache" parent="ARVROrigin/ARVRCamera" instance=ExtResource( 1 )]
+
+[node name="LeftHand" type="ARVRController" parent="ARVROrigin"]
+transform = Transform( 1, 0, 0, 0, 1, 0, 0, 0, 1, -0.5, 1, -0.5 )
+
+[node name="RightHand" type="ARVRController" parent="ARVROrigin"]
+transform = Transform( 1, 0, 0, 0, 1, 0, 0, 0, 1, 0.5, 1, -0.5 )
+controller_id = 2
+
+[node name="StartXR" parent="." instance=ExtResource( 5 )]
+
+[connection signal="xr_ended" from="StartXR" to="." method="_on_StartXR_xr_ended"]
+[connection signal="xr_started" from="StartXR" to="." method="_on_StartXR_xr_started"]
--- a/addons/godot-xr-tools/xr/start_xr.gd
+++ b/addons/godot-xr-tools/xr/start_xr.gd
@@ -1,0 +1,279 @@
+tool
+class_name XRToolsStartXR
+extends Node
+
+
+## XRTools Start XR Class
+##
+## This class supports both the OpenXR and WebXR interfaces, and handles
+## the initialization of the interface as well as reporting when the user
+## starts and ends the VR session.
+##
+## For OpenXR this class also supports passthrough on compatible devices such
+## as the Meta Quest 1 and 2.
+
+
+## This signal is emitted when XR becomes active. For OpenXR this corresponds
+## with the 'openxr_focused_state' signal which occurs when the application
+## starts receiving XR input, and for WebXR this corresponds with the
+## 'session_started' signal.
+signal xr_started
+
+## This signal is emitted when XR ends. For OpenXR this corresponds with the
+## 'openxr_visible_state' state which occurs when the application has lost
+## XR input focus, and for WebXR this corresponds with the 'session_ended'
+## signal.
+signal xr_ended
+
+
+## If true, the XR interface is automatically initialized
+export var auto_initialize : bool = true
+
+## If true, the XR passthrough is enabled (OpenXR only)
+export var enable_passthrough : bool = false setget _set_enable_passthrough
+
+## Physics rate multiplier compared to HMD frame rate
+export var physics_rate_multiplier : int = 1
+
+
+## Current XR interface
+var xr_interface : ARVRInterface
+
+## XR active flag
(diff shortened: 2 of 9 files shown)


Malcolmnixon:
@dsnopek Could you review this PR. I played with it on my Quest 2 and it seems to work fine.

BastiaanOlij (approved):
lgtm, but I think @dsnopek should review too :)

dsnopek:
I attempted to test this really quick but was getting an error about first_person_controller_vr.tscn being missing? But I don't think that has anything to do with this PR.

Just from skimming the code, this looks great!

Malcolmnixon:
> I attempted to test this really quick but was getting an error about first_person_controller_vr.tscn being missing? But I don't think that has anything to do with this PR.
> 
> Just from skimming the code, this looks great!

This is because we have a dependency on the godot_openxr library for that scene. @BastiaanOlij can you think of any way of removing this dependency when it's a WebXR only situation?

dsnopek:
Or maybe xr-tools should have it's own copy of the first person controller scene? This would come up with any other XR API as well.

BastiaanOlij:
Owh yeah, using the `first_controller_person.tscn` from the plugin is a convenience and also best practice for anyone targeting OpenXR, but it is not a requirement.

For a multi platform setup one could even argue that you don't want to go this way. There is nothing stopping us from just adding a normal `ARVROrigin/ARVRCamera/ARVRController` setup to the main scene and reproducing the initialisation script in the demo itself. It can then even be a script where a lot of overlapping things are used. You would end up with an init function like (Godot 3 pseudocode, I think this would be worth implementing in both 3 and 4):
```
var xr_interface : ARVRInterface

func _setup_for_openxr():
    ## do specific stuff for openxr
    ...

func _setup_for_webxr():
    ## do specific stuff for webxr
    ...

func _generic_setup():
    var vp = get_viewport()  
    vp.arvr = true

    ## Should obtain headset FPS in our _setup_for_xyz() functions and use that here but with a multiplier
    ## physics should really run at something like double the frame rate of the headset so lower margins become stable
    Engine.iterations_per_second = 90

func initialise() -> bool:
    xr_interface = ARVRServer.find_interface('OpenXR')
    if xr_interface and xr_interface.initialize():
        _setup_for_openxr()
        _generic_setup()
        return true

    xr_interface = ARVRServer.find_interface('WebXR')
    if xr_interface and xr_interface.initialize():
        _setup_for_webxr()
        _generic_setup()
        return true

    xr_interface = null
    return false
```

Malcolmnixon:
There are a few things the first_person_controller_vr.tscn from godot_openxr includes:
 - The first_person_controller_vr.gd which you were mentioning above and includes passthrough enabling
 - "configuration" which uses OpenXRConfig.gdns and provides:
   - View Configuration
   - Form Factor
   - Play Space Type
   - Action Sets
   - Interaction Profiles
 - Controllers using controller.gd which hides the controllers when the tracking confidence drops to none

I'm guessing the "configuration" node might be necessary, but it also can't be present in the scene if it's a WebXR only version and godot_openxr hasn't been placed in the assets folder.

I assume what triggers Godot to actually use godot_openxr is the presence of the "godot_openxr.gdnlib" with the "singleton" option enabled.

Malcolmnixon:
I had to rework this pull request fairly heavily to support WebXR without having the godot_openxr asset present. The change includes a new StartXR node which can be added to the staging scene:
image

Note that the scene now has a standard ARVROrigin rather than using the godot_openxr FPController scene.

This node will detect both OpenXR and WebXR and start the appropriate interface. It has a few basic properties which we could extend later:
image

Malcolmnixon:
@dsnopek Could you try this again.

dsnopek:
Works perfectly in my testing!

Skimming the diff, the rework seems great to me too :-)

dsnopek (approved):

Malcolmnixon:
Are there any use-cases where projects would want to start XR but not use the staging infrastructure?

We could move the StartXR and the ARVROrigin/camera/controllers into the plugin staging; so staging-based projects get it for free, but non-staging projects can drop in the StartXR node.

dsnopek:
> Are there any use-cases where projects would want to start XR but not use the staging infrastructure?

Personally, I don't always use the staging infrastructure when prototyping something simple, or just testing an individual scene.

> We could move the StartXR and the ARVROrigin/camera/controllers into the plugin staging; so staging-based projects get it for free, but non-staging projects can drop in the StartXR node.

This sounds good!

Malcolmnixon:
I've updated the PR with moving the StartXR node as well as the ARVROrigin/camera/controllers into the base staging scene. This makes creating a customized staging scene significantly easier - in fact the base one will work unless the developer wishes to customize the prompt_for_continue and pause/resume logic.

The StartXR node is still available for projects which don't want to use staging.

Malcolmnixon:
Merged with Godot 4 equivalent functionality.

addons/godot-xr-tools/staging/staging.tscn
```diff
@@ -29,9 +31,27 @@ material/0 = SubResource( 3 )
...
 
 [node name="LoadingScreen" parent="." instance=ExtResource( 3 )]
-splash_screen = null
 progress = 0.0
 
 [node name="Scene" type="Spatial" parent="."]
 
 [node name="Tween" type="Tween" parent="."]
+
+[node name="ARVROrigin" type="ARVROrigin" parent="."]
```
BastiaanOlij, line 40: I wonder if we should rename the nodes to `XROrigin` and `XRCamera`, might make things more recognizable between 3 and 4? Or am I over thinking it?

addons/godot-xr-tools/xr/start_xr.gd
```diff
@@ -0,0 +1,275 @@
...
+
+
+## If true, the XR interface is automatically initialized
+export var auto_initialize : bool = true
+
+## If true, the XR passthrough is enabled (OpenXR only)
+export var enable_passthrough : bool = false setget _set_enable_passthrough
+
+## Physics rate multiplier compared to HMD frame rate
+export var physics_rate_multiplier : float = 1
```
BastiaanOlij, line 36: I wonder if we should make this an integer, a fractional multiplier may lead to jittering as physics frames are interjected in only some render frames. The most important use case here is doubling the physics frames to get better collision detection at the scales we're using.

addons/godot-xr-tools/xr/start_xr.gd
```diff
@@ -0,0 +1,275 @@
...
+	if !Engine.editor_hint and auto_initialize:
+		initialize()
+
+
+## Initialize the XR interface
+func initialize() -> bool:
+	# Check for OpenXR interface
+	xr_interface = ARVRServer.find_interface('OpenXR')
+	if xr_interface:
+		_setup_for_openxr()
```
BastiaanOlij, line 63: Come to think of it, `_setup_for_openxr` should return bool and we should change this to `return _setup_for_openxr()` so if initialization fails, we can react appropriately.

addons/godot-xr-tools/xr/start_xr.gd
```diff
@@ -0,0 +1,275 @@
...
+	if refresh_rate > 0:
+		# Report provided frame rare
+		print("OpenXR: HMD refresh rate is set to ", str(refresh_rate))
+	else:
+		# None provided, assume a standard rate
+		print("OpenXR: No refresh rate given by XR runtime")
+		refresh_rate = 144
+
+	# Pick a physics rate
+	var physics_rate := int(refresh_rate * physics_rate_multiplier + 0.5)
```
BastiaanOlij, line 140: `+0.5`?

BastiaanOlij (approved):
Just a few nitpicking things but I think this is a great change making it easier to start with a cross platform solution.

addons/godot-xr-tools/staging/staging.tscn
Malcolmnixon: It has benefits for compatibility, but it may also be a source of confusion as the developers will still have to use the ARVRxxxx names in code. If we choose to do this then I would probably make it a separate PR as we'd want to sweep through other files and comments.

addons/godot-xr-tools/xr/start_xr.gd
Malcolmnixon: Done - both functions return bool which is returned through initialize()

addons/godot-xr-tools/xr/start_xr.gd
Malcolmnixon: Done - I thought you'd mentioned 1x, 1.5x, and 2x at some point; but integer would be better.

addons/godot-xr-tools/xr/start_xr.gd
Malcolmnixon: It's a fast way of rounding positive values to integer as casting to int does a floor operation; however for clarity I've changed it to int(round(<expression>))

Issue #315, opened 2023-01-08
WebXR games show "Enter VR" if they restart


Malcolmnixon:
A common trick for simple one-scene games is to restart via ```get_tree().reload_current_scene()```

When playing through WebXR, the StartXR node doesn't check that the viewport is already in VR mode, and instead pops up the Enter VR canvas while in 3D.

BastiaanOlij:
resolved by #316.

PR #316, opened 2023-01-08
Prevent Enter VR button from showing if XR is already active.

Malcolmnixon:
This pull request fixes issue #315 by only kicking off the WebXR session initialization sequence if XR isn't already running.


dsnopek (approved):
Looks good to me!

BastiaanOlij (approved):

Issue #311, opened 2023-01-02
Prevent snap zones from allowing objects to "fall" into place


dsnopek:
Currently, pickable objects can fall (or be thrown) into a snap zone, which isn't always desirable. It looks pretty funny with the teacups in the pickable demo:

![xrtools-drop-teacup](https://user-images.githubusercontent.com/191561/210191574-14b60a98-7bc1-45d3-a084-28dbee1ca8d5.gif)

It'd be great to have a mode for snap zones that required the object to be placed directly in the snap zone!

I've been struggling with how to implement this, though. My first idea was to have the snap zone connect to the "dropped" signal on any pickable objects that enter, but that clashes with the pickable objects clearing out their collision layers when being held (at least by default).

Any ideas?

Malcolmnixon:
It's actually one part of a two-part problem which includes the initial snapped state on scene loading. The current approach is to position objects in the snap zones and hope they get proximity-snapped when the scene loads. That works for desktop, but it turns out running the demo scene above native on a quest doesn't correctly snap and the teacup falls on the table.

So the two parts to this are:
 - Support snap modes of [OnEnter, OnDrop]
 - Support NodePath for snap-on-load

The OnDrop may be something we can do with the XRToolsFunctionPickup as it maintains a list of all pickable objects in range and the snap zones are pickable (so you can grab items from them).

Essentially when the XRToolsFunctionPickup drops an object it could search its list of pickup areas for anyone with an on_dropped_near method and let them grab it.

dsnopek:
> The OnDrop may be something we can do with the XRToolsFunctionPickup as it maintains a list of all pickable objects in range and the snap zones are pickable (so you can grab items from them).
>
> Essentially when the XRToolsFunctionPickup drops an object it could search its list of pickup areas for anyone with an on_dropped_near method and let them grab it.

Aha! This sounds way better than any of the things I've been trying so far :-)

Malcolmnixon:
Another thought on detecting pickables dropped in snap zones is that the snap zones could subscribe to the pickables "dropped" signal when the pickable enters the snap zones area, and unsubscribes when it leaves.

BastiaanOlij:
Resolved by #319.

PR #319, opened 2023-01-08
Added Snap-zone modes to XRToolsSnapZone

Malcolmnixon:
This pull request implements issue #311 by:
 - Modified XRToolsGrabPointSnap so it only allows grabbing from snap-points
 - Modified XRToolsSnapZone to have a snap_mode property
 - Modified XRToolsSnapZone so it snaps dropped objects in range when snap_mode == DROPPED
 - Modified XRToolsSnapZone so it snaps objects in range when snap_mode == RANGE
 - Modified the demo scenes to support the new modes - the saucer is the only RANGE snap-zone


addons/godot-xr-tools/objects/snap_zone.gd
```diff
@@ -208,3 +250,30 @@ func _set_grab_distance(new_value: float) -> void:
...
+# Called when the snap mode property has been modified
+func _set_snap_mode(new_value: int) -> void:
+	snap_mode = new_value
+	if is_inside_tree():
+		_update_snap_mode()
+
+
+# Handle changes to the snap mode
+func _update_snap_mode() -> void:
+	set_process(snap_mode == SnapMode.RANGE)
```
BastiaanOlij, line 264: We should loop through the objects currently in range and either connect or disconnect our dropped signal when applicable.

If someone changes the snap mode in runtime we could end up with it not working properly

BastiaanOlij (approved):
lgtm outside of that one nitpick.

addons/godot-xr-tools/objects/snap_zone.gd
Malcolmnixon: Updated to handle dynamic switching of modes correctly.

PR #321, opened 2023-01-08
Make Demos show fancier

pietru2004:
I added code so if you hide demo in editor it will hide it and reposition active demos instead of keeping them on positions. Teleporters also got is active checkbox/property.

### DIFF
--- a/scenes/main_menu/main_menu_level.gd
+++ b/scenes/main_menu/main_menu_level.gd
@@ -4,10 +4,11 @@ extends "../scene_base.gd"
 func _update_demo_positions() -> void:
 	var count = 0#$Demos.get_child_count()
 	var vis_chd := []
-	for _c in $Demos.get_children():
-		if _c.visible:
+	for _tp in $Demos.get_children():
+		_tp.active=_tp.visible
+		if _tp.visible:
 			count+=1
-			vis_chd.append(_c)
+			vis_chd.append(_tp)
 	if count > 1:
 		var angle = 2.0 * PI / count
 		for i in count:
--- a/scenes/main_menu/main_menu_level.gd
+++ b/scenes/main_menu/main_menu_level.gd
@@ -6,10 +6,7 @@ func _update_demo_positions() -> void:
 	var visible_children := []
 	for teleporter in $Demos.get_children():
 		teleporter.active=teleporter.visible
-		if !Engine.is_editor_hint():
-			for child in teleporter.get_node("TeleportBody").get_children():
-				if child is CollisionShape3D:
-					child.disabled=!teleporter.visible
+		teleporter.set_collision_disabled(!teleporter.visible)
 		if teleporter.visible:
 			count+=1
 			visible_children.append(teleporter)
--- a/scenes/main_menu/main_menu_level.gd
+++ b/scenes/main_menu/main_menu_level.gd
@@ -2,13 +2,13 @@
 extends "../scene_base.gd"
 
 func _update_demo_positions() -> void:
-	var count = 0#$Demos.get_child_count()
-	var vis_chd := []
-	for _tp in $Demos.get_children():
-		_tp.active=_tp.visible
-		if _tp.visible:
+	var count = 0
+	var visible_children := []
+	for teleporter in $Demos.get_children():
+		teleporter.active=teleporter.visible
+		if teleporter.visible:
 			count+=1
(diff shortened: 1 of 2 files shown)


scenes/main_menu/main_menu_level.gd
```diff
@@ -2,20 +2,29 @@
 extends "../scene_base.gd"
 
 func _update_demo_positions() -> void:
-	var count = $Demos.get_child_count()
+	var count = 0#$Demos.get_child_count()
```
BastiaanOlij, line 5: I'd remove the comment

scenes/main_menu/main_menu_level.gd
```diff
@@ -2,20 +2,29 @@
...
+	var vis_chd := []
```
BastiaanOlij, line 6: I'm not a big fan of shorthand unless names get really long, I would rename this to `visible_children`.

scenes/main_menu/main_menu_level.gd
```diff
@@ -2,20 +2,29 @@
...
+	for _tp in $Demos.get_children():
```
BastiaanOlij, line 7: Kind of the same with `_tp`, also the `_` prefix is a convention to indicate a private variable, not really applicable for local variables. I would just call this `teleporter`

scenes/main_menu/main_menu_level.gd
```diff
@@ -2,20 +2,29 @@
...
 
-			$Demos.get_child(i).transform = t
+			vis_chd[i].transform = t
 
 
 func _ready():
 	super._ready()
 	_update_demo_positions()
+	
+	for _c in $Demos.get_children():
```
BastiaanOlij, line 26: Same here, just rename `_c` to `child` or keep it consistent and all it `teleporter`

BastiaanOlij (approved):
Just a few nitpicking things in naming. Looks good!

scenes/main_menu/main_menu_level.gd
pietru2004: To be fair it was _c in begining, but later I added `teleport.active`

scenes/main_menu/main_menu_level.gd
pietru2004: Ok, I gona do it, but I don't think phone github will allow me so probably I gona sit to it in 8H

scenes/main_menu/main_menu_level.gd
pietru2004: Also I think I gona go with world `teleporter` cause I think it will be faster to understand when somebody copies repo/downloads demos

Malcolmnixon:
A hidden teleporter still has it's physics collision enabled, and it's possible to walk over to an invisible teleporter and stand on its still-physical base.

pietru2004:
@Malcolmnixon is that good?
obraz

pietru2004:
I should probably add check if it is not a tool...

Malcolmnixon:
It seems like it should work (and supports both Godot 3 and 4), but I need to check when I get home this evening - there's some odd verbiage in the documentation about CollisionShape3D - _"IMPORTANT: this is an Editor-only helper to create shapes"_ but it doesn't exactly say what the ramifications are.

BastiaanOlij:
> It seems like it should work (and supports both Godot 3 and 4), but I need to check when I get home this evening - there's some odd verbiage in the documentation about CollisionShape3D - _"IMPORTANT: this is an Editor-only helper to create shapes"_ but it doesn't exactly say what the ramifications are.

It likely means this class is only included in editor builds, could break on runtimes

BastiaanOlij:
> @Malcolmnixon is that good? obraz

Not a fan, should expose a "disabled" property on the teleporter itself that nicely disables it's own collision shape.

Malcolmnixon:
> Not a fan, should expose a "disabled" property on the teleporter itself that nicely disables it's own collision shape.

I agree - the teleporter should take on some of the responsibility. I'm also planning on moving the teleport-ring logic to its own node at some point - other scenes may want to have their own set of teleporters, or we may want to group similar functionality into different "areas" of the main scene - for example a ring dedicated to "movements" and a different one for "interactions".

pietru2004:
ok I moved collision disabling code to other place

Malcolmnixon:
Looks good to me and works correctly with the collision shapes being removed.

PR #322, opened 2023-01-09
Flapping wings. Add vertical impulse when flapping arms

surreal6:
This PR adds a new option to the movement_glide script. It adds a vertical impulse when flapping your arms. The impulse is relative to the downwards velocity of the controllers.

It has a boolean 'wings_impulse' to activate the feature (default is set to false)
It also has an exported 'wings_force' variable to multiply the vertical impulse.

Also added some floating rings to the glide scene to play fly through them.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -13,5 +13,6 @@ Other people who have helped out by submitting fixes, enhancements, etc are:
 - [Sam Sarette](https://github.com/lunarcloud)
 - [Henodude](https://github.com/Henodude)
 - [Miodrag Sejic](https://github.com/DigitalN8m4r3)
+- [Carlos Padial](https://github.com/surreal6)
 
 Want to be on this list? We would love your help.
--- a/addons/godot-xr-tools/functions/movement_glide.gd
+++ b/addons/godot-xr-tools/functions/movement_glide.gd
@@ -28,13 +28,6 @@ signal player_glide_end
 ## Signal invoked when the player flaps
 signal player_flapped
 
-## flap activated (when both controllers are near the ARVRCamera height)
-var flap_armed : bool = false
-
-## last controllers position to calculate flapping velocity
-var last_local_left_position : Vector3
-var last_local_right_position : Vector3
-
 ## Movement provider order
 export var order : int = 35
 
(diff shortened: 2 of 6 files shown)


Malcolmnixon:
I *REALLY* like it - it's definitely lots of fun, if not tiring on the arms - and I made a game last year where you could turn into a bird, but I only had the gliding ;)

I think you need to add a "flapped" signal which you fire when you detect the player flap their arms - a game would want to subscribe to it and play a bird-flap sound to go along with the arm movements. To do that without glitching, you'd need to improve the flap detection - possibly reporting when both hands are moving down quickly (greater than 0.3 m/s) and then require both hands to go up before re-arming.

surreal6:
ok, thats a good idea. I'll try.

Maybe i can activate the wings only when both controllers are above the head.

One question, where do i find the gdlint specification for this repo? Do you know if there is an eslint plugin for gdscript in visual studio?

Malcolmnixon:
The gdlint project is at https://github.com/Scony/godot-gdscript-toolkit and has the installation and running instructions.

It enforces the coding style from https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html

You can also see the issues list in this pull request page, although it's obviously better to fix them before pushing.

surreal6:
i have the logic working :)  

need to check the lint stuff. Will do it in the next days.

Also, can update the docs for the glide movement in this same PR if you prefer.

BastiaanOlij:
@surreal6 docs are in a separate branch so it'll have to be a separate PR.

BastiaanOlij (approved):
Cool addition, can be merged after lint issues are resolved and if Malcolm hasn't any further wishes for improvements.

Also don't forget to add your name to contributors.md

Malcolmnixon:
I need to do some additional checks on this. Originally glide only fired when you were "falling", so for example a jump-pad would throw you into the air, but you couldn't glide until you reached the maximum height and started falling back down.

You removed the falling check (because flapping involves moving the player up), however we may want to restore that condition if flapping is disabled.

I may need to add a jump-pad or two to the gliding demo scene to verify functionality. Also the gliding demo should probably prevent the player from flying too far (disable flapping if the player is above the top of the cliffs) otherwise the player will be able to fly/glide off the edge of the world.

surreal6:
Check it and tell me! 

Please note I didn't remove the falling check, but split it in two different checks:

```
if player_body.on_ground:
		_set_gliding(false)
		return
var wings_impulse_velocity := 0.0

if wings_impulse:
		** here i calculate the wings impulse
		** and after that: 

if vertical_velocity >= glide_min_fall_speed && wings_impulse_velocity == 0.0:
		_set_gliding(false)
		return
```

With this, if you jump and make a strong flap at the same time, you can take off, but if you are not flapping, it works the same as before (as i understand, maybe i'm wrong)

To put a limit to the flying altitude is a good idea.

Malcolmnixon:
There still might be some future enhancements that can be added - for example supporting roll-turning while flapping - currently the moment the player flaps, the player starts moving up and so the roll-turning shuts off.

Issue #233, opened 2022-10-27
Movement Dead zone


pietru2004:
Would there be possibility to have developer set movement deadzone? I know it can be set in for example in steam, but I find setting deadzone at game level to .1 / .2 as handy thing.

_Not everybody will know that it is possible to set joystick deadzone in settings - I didn't know till recently when browsing throu steam vr settings_

I had random movement at times without using direct movement so I made my small override
```
# Apply forwards/backwards ground control
	if abs(_controller.get_joystick_axis(1))>.2:

# Apply left/right ground control
	if strafe and abs(_controller.get_joystick_axis(0))>.2:
```

PR #323, opened 2023-01-09
issue #233: add deadzone to direct movement joystick inputs

surreal6:
I find #233 tagged as 'good first issue' for newcomers so i made this PR

Not sure if it will be better to put the deadzone values in the project settings as i see in the snap turning logic...

let me know if i can improve this


Malcolmnixon:
I think we're going to end up moving **most** of the movement properties into the project settings soon. It makes more sense to have them there than scattered around numerous scenes and possibly forgetting to update one. Also it puts the settings in a single area which helps the developer manage them all.

BastiaanOlij:
Yes, agreed these should become project settings similar to our grip threshold setting.

BastiaanOlij:
Discussed in XR meeting, adding deadzones and other thresholds to XR tools seems warrented. There is likelyhood that things will become available in OpenXR itself to set deadzones and threshold in the action map but that is an unknown at this point in time, both timing wise and what that functionality may be.

We do suggest that instead of these being export variables that they become project settings similar to the grip threshold we added not long ago.

@surreal6 would you mind making that change?

surreal6:
Of course 👌 I can make the changes

surreal6:
I'll be out for two weeks, so can't do the update until then , sorry

surreal6:
Hi, here it is. Not sure about the range hint values, let me know what you think.

Malcolmnixon:
@BastiaanOlij There are two questions I have with implementing dead-zone...

# Mapping of the outputs
image

When applying a dead-zone, do we want to just "remove" any inputs less than the dead-zone, so for example given a dead-zone of 0.2; an input of 0.19 is mapped to 0 (nothing) and an input of 0.21 is mapped to 0.21 resulting in an inability to provide small values. Conversely an alternate approach is to "rescale" the inputs, so again for a dead-zone of 0.2; an input of 0.19 is mapped to 0 (nothing) and an input of 0.21 is mapped to 0.0125 `(0.21 - 0.2) / 0.8`, so small values are still possible.

# Implemented in all providers
If we want this implemented for all providers, should we move the logic into a helper function in xr_tools.gd rather than reimplementing the logic in all providers.

surreal6:
#468  solved this and can be closed

PR #324, opened 2023-01-10
Added render_target_size_multiplier property to StartXR node.

Malcolmnixon:
This pull request adds the render_target_size_multiplier option to the StartXR node. Additionally it prints the OpenXR system name which may be useful for debugging.


BastiaanOlij (approved):
I guess we could add this though I've always been on the fence when it comes to this option. There is a reason the XR runtime suggest the optimal resolution to run on not to mention that runtimes such as SteamVR already have options for this that end up contradicting the option in Godot.

That said, we offer it as a feature so I guess we should expose it then.

PR #326, opened 2023-01-10
Remove strict class type in FunctionPickup for more flexibility in how controllers are used

goatchurchprime:
In trying to get my own networking avatar code to work with the base toolkit again, I hit a blocker in the over-requirement of the FunctionPickup feature to point to a genuine ARVRController object, rather than a node that has the methods and signals it needs to work (ie duck-typing).

image

It is necessary to attach the FunctionPickup feature to a node whose transform position is not directly set by the hardware controller in order to implement features such as (1) physics hands that don't go through walls, (2) heavy hands that droop when carrying a large object, and (3) to make hand tracking and controller tracking inputs interchangeable by mapping their transforms to the same GripArea.

### DIFF
--- a/addons/godot-xr-tools/functions/function_pickup.gd
+++ b/addons/godot-xr-tools/functions/function_pickup.gd
@@ -202,7 +202,7 @@ static func find_right(node : Node) -> XRToolsFunctionPickup:
 
 
 ## Get the [ARVRController] driving this pickup.
-func get_controller() -> ARVRController:
+func get_controller() -> Node:
 	return _controller
 
 
--- a/addons/godot-xr-tools/functions/movement_turn.gd
+++ b/addons/godot-xr-tools/functions/movement_turn.gd
@@ -57,7 +57,7 @@ func physics_movement(delta: float, player_body: XRToolsPlayerBody, _disabled: b
 		deadzone = XRTools.get_snap_turning_deadzone()
 
 	# Read the left/right joystick axis
-	var left_right := _controller.get_joystick_axis(XRTools.Axis.VR_PRIMARY_X_AXIS)
+	var left_right = _controller.get_joystick_axis(XRTools.Axis.VR_PRIMARY_X_AXIS)
 	if abs(left_right) <= deadzone:
 		# Not turning
 		_turn_step = 0.0
--- a/addons/godot-xr-tools/misc/arvr_helpers.gd
+++ b/addons/godot-xr-tools/misc/arvr_helpers.gd
@@ -79,18 +79,24 @@ static func get_arvr_camera(node: Node, path: NodePath = NodePath("")) -> ARVRCa
 ## The caller may provide an optional path (relative to the node) to the
 ## [ARVRController] to support out-of-tree searches.
 ##
+
 ## The search is performed assuming the node is under the [ARVRController].
-static func get_arvr_controller(node: Node, path: NodePath = NodePath("")) -> ARVRController:
-	var controller: ARVRController
+static func get_arvr_controller(node: Node, path: NodePath = NodePath("")) -> Node:
+	var controller: Node
 
 	# Try using the node path first
 	if path:
-		controller = node.get_node(path) as ARVRController
-		if controller:
+		controller = node.get_node(path)
+		if controller and controller.has_signal("button_pressed"):
 			return controller
 
 	# Search up from the node for the controller
-	return XRTools.find_ancestor(node, "*", "ARVRController") as ARVRController
+	# XRTools.find_ancestor(node, "*", "ARVRController") as ARVRController
+	while node:
+		if node.has_signal("button_pressed"):
+			break
+		node = node.get_parent()
+	return node
 
 ## Find the Left Hand Controller from a player node and an optional path
 static func get_left_controller(node: Node, path: NodePath = NodePath("")) -> ARVRController:


Malcolmnixon:
With Godot 4 strong-typing is significantly faster than duck-typing, which is why we have been switching the code-base over to strong types everywhere.

There are large portions of the code-base that assume when they call "get_arvr_controller" that they're really given an ARVRController because they're going to do many other things with them:
 - drive the ```rumble``` property
 - query the ```controller_id```
 - use ```get_is_active()``` to verify the controller is on
 - use ```get_joystick_axis()``` to read the joystick

As such just using a duck-type check for the "button_pressed" signal is going to potentially result in all these other objects being given something that doesn't meet their expected API contract.

We have been planning on adding physics-blocked hands for a while. The approach we were considering was to have a version of the hands which did ```set_as_toplevel(true)``` then attempted to move to its parent using move_and_slide. The FunctionPickup could then be childed to the hand object, so it would follow the physics hand, but still find its ARVRController as an ancestor.

goatchurchprime:
That comes to the same thing.  We need to be able to tell the Movement functions to get their button inputs from one node (the ARVRController), and the transform position from a different Node.  

Could we do this sneakily with the following remotr transform node:
  ARVRController/Physics_blocked_hand/RemoteTransform -> ARVRController/FunctionPickup

Then the FunctionPickup position is being set from the RemoteTransform from another object, but it still looks to the ARVRController parent for its controls?

This, however, doesn't help with trying to use the same FunctionPickup node for both the Controller tracking and the Hand tracking, but maybe we don't want that.  Also, I guess we can use disabled ARVRControllers in the enemies or other player avatars if we want to use the same FunctionPickup node to pick up objects with their appendages while preserving the strong typing of the variables.

BastiaanOlij:
@goatchurchprime Malcolm is pointing to the "is_top_level" switch, it's hidden but there in Godot 3, it is now exposed as a tickbox in Godot 4. It allows a node to be positioned independent of it's parent node and there is thus no need for remote transforms or nodes in other trees as far as XRTools is concerned. Combine that with the changes we've been making that allow functions to be a child of a child of the controller node you have a solution.

So your XR tree would look like this:
```
- XROrigin3D
  - XRController3D
    - XRPhysicsHand (is_top_level = true)
      - XRPickupFunction
```
Now the physics hand attempts to follow the controller but can be stopped by a wall. The pickup function works on this location.
If you have an avatar managed by a separate tree you can obviously add remote transforms to this to make that work, but as far as the XRTools nodes are concerned, it's all context based.

This is an important thing to us, and to Godot in general. As you may remember we started out with many functions which had a number of node paths as properties but it adds a lot of complexity and makes the toolkit much harder to understand and use, especially when talking about optional functionality that isn't immediately clear to the user, is optional. 

Simply consistently making a node work based on placing it in a logical position (i.e. as a child of the controller its related to) is far more natural, so we're pretty adamant we follow that design guideline as much as we can.

BastiaanOlij:
> This, however, doesn't help with trying to use the same FunctionPickup node for both the Controller tracking and the Hand tracking, but maybe we don't want that. 

This makes less sense to me, using the same pickup node for both hands would make the code far more difficult to have it work sensible, especially as the pickup function also interacts with grab points to position objects when they are held. 
Suddenly you need to add a lot of extra control logic so it can handle multiple held objects and track which hand is being used for what.
Having this context driven by having a pickup function for each hand childed to the hand it functions for, is a major simplification code wise.

> Also, I guess we can use disabled ARVRControllers in the enemies or other player avatars if we want to use the same FunctionPickup node to pick up objects with their appendages while preserving the strong typing of the variables.

This one is more difficult, because it can differ very much from implementation to implementation. At face value I think you'll find you're trying to fit a square peg in a round hole. Enemy AI pickup logic functions very differently to player logic. There is some overlap in needed a detection area and a mounting point, but how the pickup action is trigger and handled is structurally different.
It would very likely be far easier to create a purpose build enemy AI pickup function then to try and repurpose the XR tools one for the job.

Malcolmnixon:
I'm going to have to familiarize myself with the hand-tracking logic (I personally hate using empty hands in VR). Godot-xr-tools is currently fairly strongly tied to controller input in order to read all the useful inputs (move/turn/strafe/grip/trigger/jump/sprint/etc.) and if the hand-tracking logic doesn't drive the controllers correctly then things will start to fail badly.

The -1/+1 grip-range sounds like an issue that was fixed for real controllers back in February of 2022 (https://github.com/GodotVR/godot_openxr/issues/157) but apparently the hand-tracked input wasn't flagged as having the same problem. I would suggest filing a bug in godot_openxr for that issue.

I think it's also worth-while considering a new demo scene in godot-xr-tools-demo for testing hand-tracking. It may involve more work as most of the scenes won't function correctly with hand-tracking so we'd probably want the player blocked from entering them unless they're holding the real controllers.

BastiaanOlij:
Controller ids 3 and 4 are Quest specific which is a PITA. They are how the hand aim extension on the Quest is exposed which is Metas alternative solution to hand tracking not being part of the action map system. I want to be very careful doing anything in XR tools that relies on it as any such solution isn't portable and I rather wait on an official solution in OpenXR that addresses that issue properly. This is also why this extension currently is not implemented in Godot 4.

That we're exposing the additional poses through a GDNS node in Godot 3 is also a PITA, this is solved properly in Godot 4. Here we're going to have to make a choice whether we put an interim solution in for Godot 3, or just accept this is a gap that won't be filled until Godot 4.

One solution that would possibly fit well in what we're trying to accomplish is that in the `_ready` function we can detect if our parent is an `OpenXRPose` node and if so, find the `ARVRController` with the same id.

BastiaanOlij:
Thinking of closing this PR as outdated.

The controller 3/4 issue was changed in Godot 4.x

The pickup node still needs to be part of the XRController3D branch but doesn't need to be a direct child anymore. What is a common implementation now is to add a child to XRController3D that applies further constraints to the movement, be it stopping the movement from going through static colliders or restricting the movement by IK. Then XR Tools functions can be childed to those.

An avatar can thus be separate but be part of limiting the positioning of this helper node.

However this approach has only been tested for collisions, not yet for IK restrictions.

goatchurchprime:
What happens when we want NPCs or robot arms to pick up and move objects around the scene?  

If the FunctionPickup feature was generalized so that it could be placed under any node, then it wouldn't take much work to have this capability.  And a lot of the FunctionPickup implementation is in the PickableBody node.

goatchurchprime:
Hadn't spotted this is a PR, not an issue, so it's probably wildly out of date.  I'll look again at generalizing FunctionPickup so that the machinery could be used by NPCs, not just by the player's arms.

PR #330, opened 2023-01-12
Single swapping function pointer3

goatchurchprime:
Changes to the implementation of FunctionPointer so that when it is placed under the ARVROrigin instead of under a single hand it swaps its position between which hand was most recently active.

An implementation of this feature was done externally in the PointerDemo.  Now that unnecessary code can be deleted and everyone can use this feature without any extra coding just by putting one FunctionPointer under the ARVROrigin.


BastiaanOlij (approved):
Nice one! I like this approach

BastiaanOlij:
@Malcolmnixon can you have a look too whether you're happy with this and merge if you are?

Malcolmnixon:
When I run the demo and enter the PointerDemo scene, the laser-pointer is floating detached away from the hands. I need to squeeze trigger on one of the controllers before it actually connects to the hands.

goatchurchprime:
I'll sort that out.  I think I can also make it work automatically for those ARVRController.controller_id 4 and 5 automatically, just as an extra feature.

Malcolmnixon:
This now works fine. Excellent work!

PR #331, opened 2023-01-13
single snap turn when step turn delay is zero

goatchurchprime:
This replaces https://github.com/GodotVR/godot-xr-tools/pull/290 because I couldn't 
work out how to squash my commits down into one


Malcolmnixon (approved):
This looks good - it's the squashed version of the previously approved changes.

PR #334, opened 2023-01-16
Function pointer swapping with hand pointers

goatchurchprime:
This is a further upgrade on the function_pointer so it can use the ARVRController.controller_id=3 or 4 for hand poses.

It also responds to a thumb-index pinch gesture from the hand tracking.

Run the demo scene and go into the pointer_demo.  Put down the controllers and wait for hands tracking to start. Then the pointer will follow your hands.  

The posed pointer based on the hand tracking is not the same as the hand position, because it is partly controlled by the direction of your arm in order to stabilize its orientation and make it easier to aim at UI components in spite of the noise in the hand-tracking.  As such, it can work independently of actually needing to draw hands.  Think of it as rendering the mouse position on a computer screen while not paying attention to the position of the mouse on the table.  Your hand is the mouse on the table (its exact position doesn't matter) driving the mouse pointer on the screen (or in VR).

This to be applied after #330 (if at all!)


BastiaanOlij:
I'd leave this out for now, as mentioned before the 3+4 controller stuff is Quest only and I'm still hoping that will get implemented very differently in the future.

I rather not have a bunch of code pollution for something we need to undo/redo while we're not supporting hand tracking in XR tools atm to begin with.

goatchurchprime:
This feature already comes with https://github.com/Godot-Dojo/Godot-XR-AH

PR #335, opened 2023-01-17
Hand tracking on the menu scene

goatchurchprime:
This is an ugly PR to implements everything I know about hand-tracking on the godot-xr-tools models, implemented in a single Function_Handtracking scene.  I have hacked the feature into the main_menu_level scene for now to save making a new demo.  It could probably go into the poke demo if I implemented the poke feature onto the hand.

The code works by setting up OpenXRPose and OpenXRSkeleton nodes for each hand and then [in the _physics_process] copying out the joint_transform arrays of the 26 joint positions of each hand which are all relative to the ARVROrigin.  This is nothing more than reverse-engineering the [OpenXRSkeleton code](https://github.com/GodotVR/godot_openxr/blob/master/src/gdclasses/OpenXRSkeleton.cpp#L94) to recover the transforms provided by the OpenXR interface.

These joint positions are then transformed to fit the skeleton of the fullglove_low hand model, which was created in Blender.  One of the main problems that prevents us from ever using OpenXRSkeleton directly is that Blender aligns bones along the Y axis, while OpenXR aligns them along -Z.

There's always going to be a lot of calculations to do here to get the any model to conform the the hand tracking because nobody produces models which match the OpenXR specification.  However, given that VRChat has made a reasonable go of mapping hand tracking to almost any avatar form, we ought to aim for a general solution as well to make it easy for people to do their thing.

image

### DIFF
--- a/addons/godot-xr-tools/functions/function_handtracking.gd
+++ b/addons/godot-xr-tools/functions/function_handtracking.gd
@@ -1,0 +1,272 @@
+extends Spatial
+class_name XRToolsFunctionHandtracking, "res://addons/godot-xr-tools/editor/icons/hand.svg"
+
+var specialist_openxr_gdns_script_loaded = false
+
+var palm_joint_confidence_L = TRACKING_CONFIDENCE_NOT_APPLICABLE
+var palm_joint_confidence_R = TRACKING_CONFIDENCE_NOT_APPLICABLE
+var joint_transforms_L = [ ]
+var joint_transforms_R = [ ]
+
+var LeftHand : XRToolsHand
+var RightHand : XRToolsHand
+
+var gxtlefthandrestdata = null
+var gxtrighthandrestdata = null
+
+onready var _controller_left_node := ARVRHelpers.get_left_controller(self)
+onready var _controller_right_node := ARVRHelpers.get_right_controller(self)
+
+# controls used to implement crude MovementDirect in hand-tracking mode
+onready var _handpointer_left_node := ARVRHelpers._get_controller(self, 
+					"ARVRController3", 3, NodePath(""))
+onready var _handpointer_right_node := ARVRHelpers._get_controller(self, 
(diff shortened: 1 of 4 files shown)


BastiaanOlij:
I want to hold off with support in XR tools around this.

First, your solution will only work on the Quest, it will fail on other platforms. You're using several Quest only features such as relying on controllers 3+4. 
Second, there are other cross platform issues in OpenXR that need to be dealt with for this logic to be consistent and usable, some of which require features only added in Godot 4 or that will only be addressed in Godot 4 (like skeleton retargeting).
Third, it is likely there will be improved non-vendor specific extensions added to OpenXR in the near future that improve this situation and allow us to create a true closs platform implementation of this feature.

This is why in XR tools we've opted to only support our own hand system that works on trigger and grip input as those are guaranteed to work cross platform.

Sadly the current state of OpenXR is that any solution involving hand tracking require platform specific implementations.

That all said, it is good to have something working for those people who want things right now, but I just don't believe they belong in the toolkit if in the near future we're going to radically change the implementation as OpenXR matures on this point.

goatchurchprime:
You must be referring to these two articles, which I hadn't seen.
https://godotengine.org/article/animation-retargeting-in-godot-4-0/
https://godotengine.org/article/animation-data-redesign-40/

I'm archiving this PR so it's not a distraction.  Hopefully it can still be discovered by anyone who *unadvisedly* tries to get hand-tracking to work in the current framework before the a new framework replaces it.

BastiaanOlij:
Thanks @goatchurchprime , discoverability of specific features is definitely something we need to figure out. Just because we deem something not suitable to be included in the core project (yet), doesn't mean it isn't worthy software that has its uses.

For me digitals avatar project is another great example of this. Its not generic enough for a generic toolkit, but it's an amazing implementation that deserves attention and for people to find and use it.

PR #339, opened 2023-01-23
Added physics layers documentation

Malcolmnixon:
Added physics layers documentation
Added player body documentation
Added wall walk documentation
Added snap zone documentation
Added reference to physics layers on pickup and pickable documentation
Updated README and LICENSE for godot-xr-tools rather than for the Jekyll template


BastiaanOlij (approved):
We may still need to have the license for the theme we're using somewhere, but yeah the license in the root should be ours. 

Other than that, good to go

Issue #345, opened 2023-01-27
Staging cannot reload the current scene


Malcolmnixon:
A scene loaded by the staging system cannot currently ask for itself to be reloaded (e.g. to reset itself). It's also a bit of a pain emitting the scene-control signals by string name. Instead XRToolsSceneBase should have methods for:
 - exit()
 - load(p_scene_path)
 - reset()

Malcolmnixon:
Resolved and demonstrated in godot-xr-template project.

PR #347, opened 2023-01-27
Added ability to reset current scene

Malcolmnixon:
This method implements feature request #345 by:
 - Adding the ability to reset the current scene
 - Adding exit_to_main_menu(), load_scene() and reset_scene() methods

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,5 +1,5 @@
 # 3.3.0 (Development)
-
+- Added reset-scene and scene-control functions to scene-base
 
 # 3.2.0
 - Minimum supported Godot version set to 3.5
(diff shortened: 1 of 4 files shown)


BastiaanOlij:
I'm unsure of the reset scene, reloading the same scene (most of which will be cached) is a nice easy way to force a reset but it's also fairly wasteful.

When you implement a scene and know you need a reset feature for that scene, it can be much more efficient to do so within the scene itself, never involving staging. In this scenario you may also wish to react very differently to a reset signal, which in this case would let interested parties know that your scene has been reset.

So yeah on the fence about that one.

Having convenience functions that trigger the signals being emitted, that makes a lot of sense. 
I'm worried about renaming the signals being breaking changes for anyone who has already been using XR Tools for awhile. It might be a little late for that.

BastiaanOlij (approved):
after discussing this earlier today, with these change this looks good to me

PR #357, opened 2023-02-02
added surface footstep - system

DigitalN8m4r3:
added surface_audio - a node that can be placed inside any static body and once you stand on it, it starts playing the previously defined footstep

added movement_footstep - a movement provider that controls the footsteps it plays in combination with the surface audio plays the corresponding footstep soundfile

added footstep_holder - holds all footstep sounds, this node gets instanced under the player_body/kinematic_node

added randomAudio and randomAudioStream - these .gd scripts are responsible for the variation in the audio of the footsteps, they play the audio files randomly and change the pitch_scale randomly


DigitalN8m4r3:
1st, thank you @Malcolmnixon for helping me inbetween, without your kind help i wouldnt have finished this.
thx

Notes:
the present footstep sounds are vocal audio files are for debugging purposes.
replacing those is not what i intend, so if someone has free open source audio for footsteps or is willing to create some from scratch, feel free to do so.

missing instructions  - if this pr should get noted to be accepted, am willing to update this with instructions

code refactoring - the code above is not perfect and i know this but it is what i was able to accomplish with my limited knowledge of .gd script and the xr tools, so whoever wants can fix my errors

Malcolmnixon (commented):
The files should all follow the snake_case_naming_convention; however a few of your files have names such as 'randomAudio.gd' or 'randomAudioStream.gd' and should be renamed.

Also script files intended for use by the author should have comments at the top describing what the file is and how someone would use it. These should use the double '##' comment style present in many of the other script files.

addons/godot-xr-tools/functions/movement_footstep.gd
```diff
@@ -0,0 +1,186 @@
...
+	return name == "XRToolsMovementFootstep" or .is_class(name)
+
+func _ready():
+	# Set as always active
+	is_active = true
+	var player_body := XRToolsPlayerBody.find_instance(self)
+	player_body.connect("player_jumped", self, "_on_player_jumped")
+
+func physics_movement(_delta: float, player_body: XRToolsPlayerBody, disabled: bool):
+	var footstep_holder := XRToolsFootstepHolder.find_instance(self)
```
Malcolmnixon, line 49: find_instance is an incredibly expensive operation. This will search for it repetitively on every physics frame, and instead the node should probably be cached in the _ready function and then used in the physics_movement function.

DigitalN8m4r3: indeed, thx for pointing that out.
it should now get var footstep_holder onready and the double call of var footstep_holder was removed as well.
previously i was calling it again inside the _on_player_jumped() function and that drained ressources too.

regarding the code comments, i added as suggested and the snake case naming, changed as suggested

Malcolmnixon:
One option for supporting any number of surface sounds is to introduce a new resource to define the surfaces. Something like:
```gdscript
tool
class_name XRToolsSurfaceType, "res://addons/godot-xr-tools/editor/icons/body.svg"
extends Resource


## XRTools Surface Type Resource
##
## This resource defines a type of surface, and the audio streams to play when
## the user lands or walks on it.


## Surface name
export var name : String = ""

## Audio stream to play when the player lands on this surface
export var hit_sound : AudioStream

## Audio streams to play when the player walks on this surface
export(Array, AudioStream) var walk_sounds : Array = []

## Walking sound minimum pitch (to randomize steps)
export(float, 0.5, 1.0) var walk_pitch_minimum : float = 0.8

## Walking sound maximum pitch (to randomize steps)
export(float, 1.0, 2.0) var walk_pitch_maximum : float = 1.2


# This method checks for configuration issues.
func _get_configuration_warning():
	# Verify the camera
	if name == "":
		return "Surface type must have a name"

	# Verify hit sound
	if !hit_sound:
		return "Surface must specify a hit audio stream"

	# Verify walk sounds
	if walk_sounds.size() == 0:
		return "Surface must specify one or more walk audio streams"

	# No configuration issues detected
	return ""
```

In the editor defined resources would look like this:
image

DigitalN8m4r3:
Thanks for reworking/ givin a way of removing the hardcoding, however right now am not able to grasp how to connect your suggestion properly and because of that am converting this to draft.
Not sure when i will be able to sort this, still thanks

DigitalN8m4r3:
thanks to @Malcolmnixon who took the time to rework the footstep system i started to be more generic and not anymore hardcoded

DigitalN8m4r3:
added instructions to the demo scene along with controls and therefore am moving to ready for review

DigitalN8m4r3:
> I ended up using this in my GGJ project (https://github.com/lunarcloud/vr-game-GGJ-2023) however play-testing immediately revealed that people want the foot-step sounds to be played more frequently when the player moves faster.

Indeed that is where gettin the sprint providers instance would come in play.
By connecting the sprint started and finished signals we might be able to get this done.
Another thing is the infinity sprint, right now you can do that endless. Stamina should be added to the sprint provider and the movement direct provider should be made available as get instance so that one can adjust the max speed to half speed to recover for sprint. Thats just an idea

Malcolmnixon:
Why not just use the ground control velocity we're already using. Specifically the step_time is decrement by delta. If it were instead decrement by delta * player_body.ground_control_velocity.length() then the rate of steps would be proportional to the players requested run speed.

BastiaanOlij:
During XR meeting, there is a lot of enthousiasm for this feature.

The question raised is if it makes sense to spin this off into a separate audio library that can be used more widely then just XR as normal games have these issues as well.

That said, we also agreed this should not stop us from finishing and merging this now, and maybe spinning this off at some later date.

Waiting on more people to review and suggest improvements.

DigitalN8m4r3:
> During XR meeting, there is a lot of enthousiasm for this feature.
> 
> The question raised is if it makes sense to spin this off into a separate audio library that can be used more widely then just XR as normal games have these issues as well.
> 
> That said, we also agreed this should not stop us from finishing and merging this now, and maybe spinning this off at some later date.
> 
> Waiting on more people to review and suggest improvements.

Spinning Off.. yeah indeed this would be a cool thing to do.
At a later point it might make sense to do this.
Right now Godot XR got almost nothing when it comes to audio and this needs to change.

Therefore ic the footsteps as the building block, next would be building a Audio Collision System with interactables support for when throwing/dropping/using/hitting with/without hands. This should close the gap we got when it comes to audio in the xr tools. 

This would come in handy with the collision hands system that i plan to get added to godot xr

Malcolmnixon:
I updated my game to use the new speed-based version. I did end up changing the `step_rate` from 0.5 to 1.0. This is essentially now how many steps per meter. It might be a good idea to rename it to 'steps_per_meter' and make it an exported property in case anyone wants to tweak it (or if people shrink the player and want the steps to change appropriately)

Malcolmnixon:
The current sounds are vocal audio files - put in because the original implementation had an exhaustive enumeration of ground types it could handle. Now that it's resource based we could get away with only having a few ground materials, and using actual audio files for them.

DigitalN8m4r3:
aye understood, just didnt got around..
will update both pr's with the above audio files and remove the placeholders

Malcolmnixon:
I think this is now ready to go in. It plays well and the sounds are good.

Malcolmnixon (approved):
Looks good

BastiaanOlij (approved):
I'm going to trust Malcolm on his review of the implementation, I'm flat out for time to actually do some proper testing.

Reading what this is intended to do and having discussed some things on discord with Malcolm about his review, I happily approve this.

This is some amazing work @DigitalN8m4r3 , thank!

PR #358, opened 2023-02-02
Update movement_sprint.gd

DigitalN8m4r3:
added missing find_instance

this might look trivial but it is good to have it just in case or like in my case where am working on a stamina provider to get the instance of the sprint and disable it that way

### DIFF
--- a/addons/godot-xr-tools/functions/movement_sprint.gd
+++ b/addons/godot-xr-tools/functions/movement_sprint.gd
@@ -162,6 +162,8 @@ func _get_configuration_warning():
 
 	# Call base class
 	return ._get_configuration_warning()
+
+
 ## Find an [XRToolsMovementSprint] node.
 ##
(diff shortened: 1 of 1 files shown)


addons/godot-xr-tools/functions/movement_sprint.gd
```diff
@@ -162,3 +162,12 @@ func _get_configuration_warning():
 
 	# Call base class
 	return ._get_configuration_warning()
+## Find an [XRToolsMovementSprint] node.
```
Malcolmnixon, line 165: Could you add a couple of blank lines between these functions - the [gdscript style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#blank-lines) suggests this style to clearly separate the functions.

Malcolmnixon (approved):
This looks good to me

Issue #359, opened 2023-02-05
An item being picked out of one snap-zone may be "stolen" by a different snap-zone


Malcolmnixon:
The scenario is as follows:
 1. Two snap-zones A and B are fairly close
 2. An object O is in snap-zone A
 3. The player attempts to grab object O
 4. Snap-zone A "drops" object O (so it can be picked up by the player)
 5. Snap-zone B detects object O being dropped in its range, and immediately grabs it

The fix is probably for snap-zones to "defer" their attempted pick-up one frame to allow any direct-grabs to finish.

Malcolmnixon:
Pull request fixes the issue

PR #361, opened 2023-02-07
Added snap-zone deferred pickup to fix snap-zone stealing.

Malcolmnixon:
This pull request fixes issue #359. The problem of snap-zone stealing is caused by snap-zones listening for objects dropped in their sphere-of-influence, and immediately picking them up. The problem is that when a player picks an object out of a snap-zone the snap-zone first drops the object. If there's a snap-zone in the vicinity that observes that initial drop then it will steal the object before the player has a chance to pick it up.

This pull request causes snap-zones to defer their pickup until idle; and then to check if the object is still available to be picked up.

### DIFF
--- a/addons/godot-xr-tools/objects/snap_zone.gd
+++ b/addons/godot-xr-tools/objects/snap_zone.gd
@@ -200,8 +200,8 @@ func _on_snap_zone_body_entered(target: Spatial) -> void:
 
 	# If this snap zone is configured to snap objects that are dropped, then
 	# start listening for the objects dropped signal
-	if snap_mode == SnapMode.DROPPED:
-		target.connect("dropped", self, "_on_target_dropped")
+	if snap_mode == SnapMode.DROPPED and target.has_signal("dropped"):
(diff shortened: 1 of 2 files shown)


addons/godot-xr-tools/objects/snap_zone.gd
```diff
@@ -281,6 +281,14 @@ func _update_snap_mode() -> void:
 
 # Called when a target in our grab area is dropped
 func _on_target_dropped(target: Spatial) -> void:
+	# The target may have been dropped out of a different snap-zone to be picked
+	# up by the player. Defer our pickup until idle, and only pick up the object
+	# if it's still available.
+	call_deferred("_try_grab_dropped", target)
```
BastiaanOlij, line 287: Instead of this, you can connect to a signal and mark it as deferred and it will do this automatically.

Malcolmnixon: Updated to defer in the connect method.

BastiaanOlij (approved):
Cool, thanks!

Issue #363, opened 2023-02-07
Quest refresh rate


BastiaanOlij:
This is a problem that in part needs to be solved in the OpenXR plugin and we're investigating things further, but even with the base problem solved, there is still an improvement to do here in `start_xr.gd`.

The Quest 2 can run at multiple refresh rates. By default it reports it is set to 90hz giving us a framerate of 90fps.
The bug we've encountered is that it is lying, it is actually running at 72hz, probably because the manifest states that both Quest 1 and Quest 2 are supported, and Quest 1 runs at a maximum of 72.

We can request the available framerates by calling `OpenXRConfig::get_available_refresh_rates()` which returns an array of refresh rates. Note that for systems that do not support the refresh rate extension this array will be empty, and for systems where the refresh rate can't be changed, we just get the current set refresh rate.

> Framerate information is not available until after the OpenXR session is ready.

We need to add a mechanism that selects a suitable refresh rate and then calls `OpenXRConfig::set_refresh_rate()`.
The Quest 2 supports 60, 72, 90 and 120, it may not be advisable to always pick 120. We may wish to add a property of the maximum refresh rate we wish to choose.

The code will thus become something like:
```
# Handle OpenXR session ready
func _on_openxr_session_begun() -> void:
	print("OpenXR: Session begun")

	# Our interface will tell us whether we should keep our render buffer in linear color space
	get_viewport().keep_3d_linear = _openxr_configuration.keep_3d_linear()

	# get supported rates
	var available_rates = _openxr_configuration.get_available_refresh_rates()
	if available_rates.size() > 0:
		# todo, remove rates higher than maximum
		
		# select the highest rate
		_openxr_configuration.set_refresh_rate(available_rates[-1])

	# increase our physics engine update speed
	var refresh_rate : float = _openxr_configuration.get_refresh_rate()
	if refresh_rate > 0:
		# Report provided frame rare
		print("OpenXR: HMD refresh rate is set to ", str(refresh_rate))
	else:
		# None provided, assume a standard rate
		print("OpenXR: No refresh rate given by XR runtime")
		refresh_rate = 144
```

DigitalN8m4r3:
just a day before i was eager to see how my overkill scenery would perform and i was surprised that it always showed 72 fps and then i compared it to the standard xr tools demo scenes and the same fps. So yeah this really needs to get sorted.
Good work on investigating the issue there.

for what its worth, a simple .gd script outputed the current fps
`extends Label3D

[...]

PR #364, opened 2023-02-08
Added frame rate control.

Malcolmnixon:
This pull request fixes issue #363 by giving the user the ability to specify a target frame-rate.

This code always tries to set the frame-rate (if supported). It picks a desired frame-rate of the users specified target frame-rate; or if it's not specified (a value of 0.0) then it will pick the frame-rate reported by the device. It then picks the closest frame-rate supported by the device and sets the device to run at that rate.

In the case where the quest lies and reports 90Hz when it's actually running at 72Hz, the value of 90Hz will be picked as the desired rate and the quest will then be set to run at that rate.


BastiaanOlij (approved):
Looks good! We probably should make a similar change to our firstpersoncontroller in the openxr plugin

PR #368, opened 2023-02-12
conversion of footstep system from 3x to 4x

DigitalN8m4r3:
this is the first attempt at converting the footstep system, currently it is broken and will need some work to get working


DigitalN8m4r3:
this is the best i could get it for now, it is on par with the godot3 version and plays the default footstep audio

issues right now, it doesnt play the other audio files, maybe the pooling isnt working right...

Malcolmnixon:
This looks good to me and is equivalent to the 3.x version

BastiaanOlij (approved):
lgtm

PR #378, opened 2023-02-19
Change player body to implement kinematic body on root level

BastiaanOlij:
This PR changes the `player_body.tscn` scene so it's root node is the KinematicBody node. It uses `set_as_toplevel` to ensure it is moved in global space and does not get its positioning from its parent.

This ensures that the player body node moves with the player and can be used to anchor objects to the player. 

There is also added logic that ensures the player body faces forward based on the camera position. This has been improved that when the player looks up or down, the up vector of the head is used giving more accuracy and stability. 

This direction is now also mixed in with a direction based on the relative positions of the hands, there is a new property called `Body Forward Mix` that gives you control of this. If set to 0 we purely use the camera forward position, if 1.0 it purely uses the forward vector of the hands. The default is 0.75 which gave decent results for me. It will never be perfect, as I've said before, we don't track the players body so any logic is guesswork.
There is a limit to this that the forward vector can not deviate more than 60 degrees to either side of the direction the player is looking at.

The pickable demo has been enhanced by adding a belt of storage snap zones to the player. These right now are unrestricted so you can do some funky things, but it is a good demonstration of attaching something useful to the player body and neatly visualises how the new forward facing logic is working.

### DIFF
--- a/addons/godot-xr-tools/player/player_body.gd
+++ b/addons/godot-xr-tools/player/player_body.gd
@@ -1,6 +1,6 @@
 tool
 class_name XRToolsPlayerBody, "res://addons/godot-xr-tools/editor/icons/body.svg"
-extends Node
+extends KinematicBody
 
 
 ## XR Tools Player Physics Body Script
@@ -59,6 +59,9 @@ export var player_height_max : float = 2.2
 ## Eyes forward offset from center of body in player_radius units
 export (float, 0.0, 1.0) var eye_forward_offset : float = 0.5
 
+## Mix factor for body orientation
+export (float, 0.0, 1.0) var body_forward_mix : float = 0.75
+
 ## Lets the player push rigid bodies
 export var push_rigid_bodies : bool = true
 
@@ -67,13 +70,6 @@ export var physics : Resource setget set_physics
 
 ## Option for specifying when ground control is allowed
 export (GroundControl) var ground_control : int = GroundControl.ON_GROUND
-
-## Collision layer for the player body
-export (int, LAYERS_3D_PHYSICS) var collision_layer : int = 1 << 19 setget set_collision_layer
-
-## Collision mask for the player body
-export (int, LAYERS_3D_PHYSICS) var collision_mask : int = 1023 setget set_collision_mask
-
 
 ## Player 3D Velocity - modifiiable by [XRToolsMovementProvider] nodes
 var velocity : Vector3 = Vector3.ZERO
@@ -141,6 +137,9 @@ var _previous_ground_local : Vector3 = Vector3.ZERO
 # Previous ground global position
 var _previous_ground_global : Vector3 = Vector3.ZERO
 
+# Player body Collision node
+var _collision_node : CollisionShape
+
 
 ## ARVROrigin node
 onready var origin_node : ARVROrigin = ARVRHelpers.get_arvr_origin(self)
@@ -148,14 +147,14 @@ onready var origin_node : ARVROrigin = ARVRHelpers.get_arvr_origin(self)
 ## ARVRCamera node
 onready var camera_node : ARVRCamera = ARVRHelpers.get_arvr_camera(self)
 
-## Player body node
-onready var kinematic_node : KinematicBody = $KinematicBody
+## Left hand ARVRController node
+onready var left_hand_node : ARVRController = ARVRHelpers.get_left_controller(self)
+
+## Right hand ARVRController node
+onready var right_hand_node : ARVRController = ARVRHelpers.get_right_controller(self)
 
 ## Default physics (if not specified by the user or the current ground)
 onready var default_physics = _guaranteed_physics()
-
-## Player body Collision node
-onready var _collision_node : CollisionShape = $KinematicBody/CollisionShape
 
 
 ## Class to sort movement providers by order
@@ -171,6 +170,20 @@ func is_class(name : String) -> bool:
 
 # Called when the node enters the scene tree for the first time.
 func _ready():
+	# Set as toplevel means our PlayerBody is positioned in global space.
+	# It is not moved when its parent moves.
+	set_as_toplevel(true)
+
+	# Create our collision shape, height will be updated later
+	var capsule = CapsuleShape.new()
+	capsule.radius = player_radius
+	capsule.height = 1.4
+	_collision_node = CollisionShape.new()
+	_collision_node.shape = capsule
+	_collision_node.transform.basis = Basis(Vector3(0.5 * PI, 0.0, 0.0))
+	_collision_node.transform.origin = Vector3(0.0, 0.8, 0.0)
+	add_child(_collision_node)
+
 	# Get the movement providers ordered by increasing order
 	_movement_providers = get_tree().get_nodes_in_group("movement_providers")
 	_movement_providers.sort_custom(SortProviderByOrder, "sort_by_order")
@@ -178,8 +191,6 @@ func _ready():
 	# Propagate defaults
 	_update_enabled()
 	_update_player_radius()
-	_update_collision_layer()
-	_update_collision_mask()
 
 func set_enabled(new_value) -> void:
 	enabled = new_value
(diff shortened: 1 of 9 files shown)


addons/godot-xr-tools/player/player_body.tscn
```diff
@@ -3,16 +3,14 @@
 [ext_resource path="res://addons/godot-xr-tools/player/player_body.gd" type="Script" id=1]
 
 [sub_resource type="CapsuleShape" id=1]
-radius = 0.4
+radius = 0.2
```
BastiaanOlij, line 6: this might be by accident

BastiaanOlij: owh this is because of the changed default on the parent, so I guess this is correct after all?

DigitalN8m4r3:
woul suggest adding zones for future inventory and storage of items
such as
- head - top (helmets/caps...)
- head - face (masks,..)
- chest - left/right (chest pockets)
- hip - left/right
- shoulder - left/right

or at least make it easy exapandable without havin to reinvent the wheel

BastiaanOlij:
> woul suggest adding zones for future inventory and storage of items such as
> 
>     * head - top (helmets/caps...)
> 
>     * head - face (masks,..)
> 
>     * chest - left/right (chest pockets)
> 
>     * hip - left/right
> 
>     * shoulder - left/right
> 
> 
> or at least make it easy exapandable without havin to reinvent the wheel

I'd be careful with this because this could make life harder for people. We have no knowledge of the avatar someone may or may not use, or how customizable the developer has made the body size. We have no knowledge if someone has support for body trackers that would require a different setup. We don't currently have logic to break the player up so we can properly handle body orientation on crouch or when the player lies down, etc.

In the end it's pretty easy to add the markers to the body for a given implementation, that would probably make a good demo instead of the two test spheres I currently have

DigitalN8m4r3:
> > woul suggest adding zones for future inventory and storage of items such as
> > ```
> > * head - top (helmets/caps...)
> > 
> > * head - face (masks,..)
> > 
> > * chest - left/right (chest pockets)
> > 
> > * hip - left/right
> > 
> > * shoulder - left/right
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or at least make it easy exapandable without havin to reinvent the wheel
> 
> I'd be careful with this because this could make life harder for people. We have no knowledge of the avatar someone may or may not use, or how customizable the developer has made the body size. We have no knowledge if someone has support for body trackers that would require a different setup. We don't currently have logic to break the player up so we can properly handle body orientation on crouch or when the player lies down, etc.
> 
> In the end it's pretty easy to add the markers to the body for a given implementation, that would probably make a good demo instead of the two test spheres I currently have

Indeed, valid points there

Guess i just got overwelmed at the prospect of gettin an actual pocket system in place. But you are totally right, it might just afterwards make things even more complicated. Thanks

addons/godot-xr-tools/player/player_body.gd
```diff
@@ -171,15 +170,27 @@ func is_class(name : String) -> bool:
 
 # Called when the node enters the scene tree for the first time.
 func _ready():
+	# Set as toplevel means our PlayerBody is positioned in global space.
+	# It is not moved when its parent moves.
+	set_as_toplevel(true)
+
+	# Create our collision shape, sizes will be update later
+	var capsule = CapsuleShape.new()
+	capsule.radius = 1.2
```
Malcolmnixon, line 179: You may want to just use `player_radius` here' although no-matter what you enter it's going to be overridden by the `_update_player_radius()` call later.

addons/godot-xr-tools/player/player_body.gd
```diff
@@ -399,6 +392,39 @@ func override_player_height(key, value: float = -1.0):
...
+		forward = up_player_plane.project(camera_forward).normalized()
+
+	if (left_hand_node and left_hand_node.get_is_active()
+		and right_hand_node and right_hand_node.get_is_active()
+		and body_forward_mix > 0.0):
+		# See if we can mix in our estimated forward vector based on controller position
+		# Note, in Godot 4.0 we should check tracker confidence
+
+		var tangent = right_hand_node.global_transform.origin - left_hand_node.global_transform.origin
+		tangent = (tangent * Vector3(1.0, 0.0, 1.0)).normalized()
```
Malcolmnixon, line 417: This math appears to assume the player is in the global Y+/UP frame; where-as the player may be walking around a sphere or on a wall.

You may want to consider something like:
```gdscript
    tangent = up_player_plane.project(tangent).normalized()
    var hands_forward = up_player_vector.cross(tangent).normalized()
```

Malcolmnixon (approved):
Looks good, but might be worth-while putting the inventory belt on the player then checking some of the scenes where the player runs around spheres or other surfaces.

teddybear082:
I played by using the sphere world, climbing world, and grappling world while playing seated and having adjusted my height up in the main menu to stress test as much as possible and seemed to work well.  One thing I find (not new to this change) is that climbing and trying to get to the top of a platform while in seated mode / height adjust up sometimes is harder than standing mode. Not sure if that makes sense or anyone else has experienced it.  I was also getting some shaking / jitter sometimes on climbing but that could have been because I had Vsync on or because of the repeating texture blocks on the walls just made it look like that. This is a great change!

DigitalN8m4r3:
just had time to test this, works pretty well so far

regarding the beltsnapzones, i know they are just for demonstrating purposes in there
just to note a few things

*they should be spread across the player body, currently they are to close to each other and this results in hard to control pickin/snapping

*pickables such as the snaptray that got their own snapzones should restrict the player from grabbing from their zones
if the snaptray is currently snapped to a zone, ideally there should be a function that blends out/ hides the childredn of the snaptray if its not being held

regarding my comments on the beltsnapzones, if there is interest, i can build upon it and submit a pr that adds an inventory backpack with some pickables along with a evenly spread beltsnapzones setup to go along

Malcolmnixon:
I've confirmed there are problems with the body position - If visible colliders are enabled and the player walks around the Origin Gravity demo then the capsule does not correspond to either the players up, or the worlds up.

addons/godot-xr-tools/player/player_body.gd
BastiaanOlij: I left that out seeing it looked weird that I properly set radius but set height to an arbitrary value while both get updated later on in the setup process to their proper values. Hence the remark `sizes will be update(d) later`, typo in that one

addons/godot-xr-tools/player/player_body.gd
BastiaanOlij: Good catch!

BastiaanOlij:
> I've confirmed there are problems with the body position - If visible colliders are enabled and the player walks around the Origin Gravity demo then the capsule does not correspond to either the players up, or the worlds up.

Thats probably due to https://github.com/GodotVR/godot-xr-tools/pull/378#discussion_r1111393937

BastiaanOlij:
> regarding my comments on the beltsnapzones, if there is interest, i can build upon it and submit a pr that adds an inventory backpack with some pickables along with a evenly spread beltsnapzones setup to go along

@DigitalN8m4r3 more than happy if you're willing to improve the demo, would be great if you could!

addons/godot-xr-tools/player/player_body.gd
BastiaanOlij: Ok, added this and the other changes you suggested below, should indeed fix the orientation issues on the spherical world

addons/godot-xr-tools/player/player_body.tscn
BastiaanOlij: Changed it and the comment :)

addons/godot-xr-tools/player/player_body.gd
BastiaanOlij: Changed it and the comment :)

Malcolmnixon:
Yup, that works beautifully now.

PR #380, opened 2023-02-22
Held objects can now detect release of action button

Malcolmnixon:
This pull request lets held objects know when the action button is released by:
 - Adding optional action_button() method for held objects.
 - Adding action_released signal on pickable objects.

This feature is demonstrated in the pickable demo scene by the green "grab cube" objects changing color while the action button is held down.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -3,6 +3,7 @@
 - Fixed snap-zones stealing objects picked out of other near-by snap-zones
 - Improved player body so it can be used to child objects to
 - Updated scene/script default physics layers to match recommendations on website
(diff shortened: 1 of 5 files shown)


BastiaanOlij:
I'm not against this but I think for more complex interactions the object being picked up should react to being picked up or dropped and register signals directly on the controller that holds it.

Especially in Godot 4 where this may result in activating relevant action sets.

Malcolmnixon:
Retiring this approach in preference of having the held objects subscribe to the controller events directly, rather than having the events routed from the controller through the pickup to the held object.

PR #382, opened 2023-02-22
Added demo for pickable object button events

Malcolmnixon:
This pull request adds a demonstration of how pickable objects can react to button presses on the controller holding them.

### DIFF
--- a/scenes/pickable_demo/objects/grab_cube.gd
+++ b/scenes/pickable_demo/objects/grab_cube.gd
@@ -1,0 +1,60 @@
+extends XRToolsPickable
+
+
+## Alternate material when button pressed
+export var alternate_material : Material
+
+
+# Original material
+var _original_material : Material
+
+# Current controller holding this object
+var _current_controller : ARVRController
+
+
+# Called when the node enters the scene tree for the first time.
+func _ready() -> void:
+	# Get the original material
+	_original_material = $MeshInstance.get_active_material(0)
+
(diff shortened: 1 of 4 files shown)


scenes/pickable_demo/objects/grab_cube.gd
```diff
@@ -0,0 +1,60 @@
...
+
+	# Listen for when this object is picked up or dropped
+	connect("picked_up", self, "_on_picked_up")
+	connect("dropped", self, "_on_dropped")
+
+
+# Called when this object is picked up
+func _on_picked_up(pickable : XRToolsPickable) -> void:
+	# Listen for button events on the associated controller
+	_current_controller = pickable.by_controller
```
BastiaanOlij, line 28: Nitpicking but I believe that we should never access member variables directly and treat them as private. They should be accessed either as a property (@export), or we should have access methods. So in this case we should introduce a `get_picked_up_by_controller()` method in XRToolsPickable.

Also seeing we're the object being picked up and we thus know that the parameter should always be self, I don't think we should be using the parameter. To the uninitiated this could suggest that we could be picked up by something other than what has picked us up.

Malcolmnixon: Added `get_picked_up_by_controller()` method and removed unnecessary `self` access via argument.

BastiaanOlij (approved):
lgtm

Issue #373, opened 2023-02-17
Missing footstep hit sound causes audio player leak


Malcolmnixon:
The footstep system doesn't check whether the hit-sound is null before acquiring a player from the idle pool; so if the sound is null the player doesn't play and is lost.

The hit-sound should be optional, as well as an optional jump-sound.

PR #392, opened 2023-02-25
Enhancements to footstep audio system

Malcolmnixon:
This pull request fixes issue #373 by avoiding attempting to play null sounds; however this PR also includes numerous other changes.

Surface audio resources now have a jump sound, and all sounds (jump, hit, and walk) are optional and can be left blank.
image

The walk speed is now a configurable property. The footsteps node now emits a footstep signal whenever it plays a footstep sound, and the audio settings have been moved to a child AudioStreamPlayer3D node called "PlayerSettings" so the footstep node doesn't have to try and replicate every audio setting (bus, mix, attenuation, etc.)
image

All reference footstep sounds have been converted to WAV files as OGG files have a fairly large delay when playing due to the effort of decompressing the audio stream.

Finally the math behind when the footstep sounds are played has been shifted so it plays at the "end" of a step rather than at the beginning.


DigitalN8m4r3:
lgtm

BastiaanOlij:
> ... the audio settings have been moved to a child AudioStreamPlayer3D node called "PlayerSettings" ...

It's not always possible but I think we should do this in more situations. It really cleans up the UI

BastiaanOlij (approved):
lgtm

BastiaanOlij:
Don't forget to change the `VERSIONS.md` document. Also we should make some documentation on this

Malcolmnixon:
Updated VERSIONS.md, and I'll work on the documentation page shortly.

Issue #348, opened 2023-01-27
Water


pietru2004:
I think there should be a demo for water diving (clear and mudy).
Swim mechanic could be made with mix of glididing flap motion and fly, but with hand actions for swim in directions.

BastiaanOlij:
Thats actually a pretty neat idea for an addition to the demo, I need to think some about this

pietru2004:
so I have some ideas for this:
-1 when player enters water check its AABB and with some kind of shader colorize players view distance
-2 Use 4.0 env. area fog to make water effect (yep the one that is having issues ATM https://github.com/godotengine/godot/issues/72331)
Make water work similar to flying, but when in water :
- a) disable gravity
- b) slow player falling speed
As for movement it could be:

-  a) using climbing to climb throu water (possibly most safe for not getting sick for those who get sick in vr)
-  b) making swim motion to move - slow hands movement causes no move, rapid hands move causes to swim (might be annoying for some people)
-  c) use direct movement or flight movement to move in 3D space of water

Starkium:
@pietru2004 this is how I was going about water in unreal engine mobile pipeline. No post processing of any kind. All color and lighting are driven through material. I'm thinking about remaking this, different and likely better in godot 4. The colors are driven from a world position height value and color curve. Ignore the top water effect. Was debugging an android vulkan thing in unreal at the time.

[Water Video](https://youtu.be/42JAHsQovw0)

pietru2004:
Looks nice

PR #393, opened 2023-02-25
Add Water Demo

pietru2004:
Water demo with usage of flight movement - Water effect works best with Forward+

Folder swim_demo

Also I did not change settings other then disabling this
obraz
and re-enabling it later after seeing this
obraz

Related - https://github.com/GodotVR/godot-xr-tools/issues/348


DigitalN8m4r3:
didnt tested it yet but there are already 2 things i can say need work

a) the naming of the transporter banner needs to say swimming
b) a 3.5 version needs to be submitted as currently we are still working on both fronts and this wont change for the time being

pietru2004:
about a) well IDK what font is used in that png, I can't make image if I don't know the font
about b) I am not sure how to implement water effect - Fog Volume into 3.5, water plane should work fine as it is just semi transparent blue-ish material with culling set to show both sides, metalness high, spectacular high and roughness at 0. So about half effect implementable. (well half effect works if you use Mobile and not Forward+)

DigitalN8m4r3:
a) just make a threewave symbol and replace the basic part with it, same i did in my footstep demo and indeed they should really state what font is to be used for the demo scenes

b) you could trigger a load of a new WorldEnvironment once the player camera enters the swimming area and reset to the one used before once player gets out of water
and in the new WorldEnvironment you could turn the fog on and change its distance and color to match approx what u got in godot4 done

pietru2004:
b) Water in my demo works by turning on and off flight. Depending on player height setting for switching fog might not look good.
In theory fog height could be used in 3.5, but what if water was in different heights - example pond upstream having waterfall into pond that is lower downstream.

pietru2004:
Also it could be possible to probably just load env, but then each pond would require having its height setted separatly, possibly creating 1pool of water - 1 env for fog

DigitalN8m4r3:
> Also it could be possible to probably just load env, but then each pond would require having its height setted separatly, possibly creating 1pool of water - 1 env for fog

you might want to take a look at this one https://godotengine.org/asset-library/asset/1567
it got the mentioned transition to water set up that way

BastiaanOlij:
I know its a PITA but this PR is changing a whole lot of unrelated import settings, probably due to changes in the new RC. We really should exclude those to cut down on the noise.

@Malcolmnixon it probably makes sense to take the 4.0-dev branch as is, and just load it up in RC5 or 6 and just make a PR that applies all these sorts of changes. I think we should stick with the latest RC as much as we can. Once stable is released we should stick to that build for awhile.

As for the demo itself, I need to find some time to try it out but I like the idea behind it.

pietru2004:
@DigitalN8m4r3 I checked out asset but well I don't think it would be best to have stuff like this in vr.
Obviously I can be wrong, but I see it would be possibly only way of making it work on 3.5. Unless they release 3.6 and add Fog Volume node.
obraz

pietru2004:
> you might want to take a look at this one https://godotengine.org/asset-library/asset/1567
> it got the mentioned transition to water set up that way

From what I see that asset uses raycast to tell if player is in water, so that would require another layer for checking is player in water

DigitalN8m4r3:
> > you might want to take a look at this one https://godotengine.org/asset-library/asset/1567
> > it got the mentioned transition to water set up that way
> 
> From what I see that asset uses raycast to tell if player is in water, so that would require another layer for checking is player in water

well you could simply use a collision mesh that triggers once u enter, and yes you are right it is not goin to be pretty in godot 3.5 but then again it at least shows how it could be done in 3.5 and it would be on par with 4.0. We all know that 3.5 lacks certain features but thats the way it is

pietru2004:
I rethinked how we cheaply can get fog effect on 3.5
obraz
I placed multiple box (they probably should have been smaller)
and made their material color alpha to 255. Then I added Distance fade
obraz

BastiaanOlij:
@pietru2004 you don't want to use distance fade or any effect that reads from the depth buffer for this. It will have serious impacts on performance on mobile VR such as Quest. It's really only suitable for PCVR.

pietru2004:
...ok...

BastiaanOlij:
Ok, had some time to take a deeper look. The concept behind this is definately good.

You've made your life a little difficult by making your changes directly on your own `4.0-dev` branch, make sure in the future that you create your own branch before you start making changes. That said, with us about to make master the 4.0 branch you probably need to do a few extra steps anyway so when the time is there I'll talk you through it.

So yeah, breaking the water up in boxes is a bad idea, it doesn't look good. Doing some sort of volumetric effect or distance fade effect will have an impact on performance as well.

You're better off to just have a plane at the top for the water with normal transparency. Making things darker as you go deeper is something you do with the texture of floor of the pond, you just make that darken based on going lower. We could also do that with a shader.

The bit that will need a bit more creativity is doing something when you're underwater. We can go oldschool and place a plane in a strategical place in front of the camera and write a nifty transparent shader based on depth.

Malcolmnixon:
The swimming mechanic works well, but rendering the water seems to be a sticking point and dependent on the type of renderer available. Let's just get a simple "water surface plane" for now so we can get the swimming demo working, and then games can decide how pretty to make the water and whether they want to use fog volumes and the vulkan renderers.

pietru2004:
Sounds good to me, still if somebody will use forward+ I want to keep fog

BastiaanOlij:
I don't think fog works right now in stereo, I still have that on my list of things to fix :)

Ok, so now that we've merged 4.0 into master we'll need to do some juggling with your PR to make it work again. 
Are you on the discord channel? if so hit me up during the week and I'll talk you through the steps to get your work cherry picked ontop of master. We'll need to squash your current commits so you have one commit, and then cherry pick the work into a new branch based on master

pietru2004:
@BastiaanOlij I am on 2 godot discords - Main and Poland One (Polish Godot Comunity). Is there Godot XR discord? Cause I can't seam to find a link.

BastiaanOlij:
@pietru2004 there is an XR channel on the main Godot discord, you just need to join the channel.

pietru2004:
Digital pm-ed me on discord with link to it - I was blind :-)

PR #397, opened 2023-02-27
basic inventory backpack

DigitalN8m4r3:
this adds a slot based inventory backpack with snap require and exclude setup

How To add the inventory to your existing scene:
select the BackpackSnapZone, located inside the xr tools objects/inventory folder, drag it under the Camera, the BackpackSnapZone is already configured to be located around your Neck/Back area, so once you grab behind your head, you will be able to pickup the holstered Backpack.

select the inventory_backpack.tscn that is located inside the objects/inventory folder of the xr tools, drag it under the Origin node (ARVROrigin/XROrigin), next select the snap_backpack.tscn from the objects/inventory folder of the pickable_demo, drag it under the InventoryBackpack Node, Next Step - select the SnapBackpack and go to Signals, click on dropped and select the InventoryBackpack Node to connect the signal.

Final Step
- select the snap_backpack.tscn scene from the pickable_demo/objects folder and dragi t on the Backpack Scene slot.
- select the BackpackInventory and drag the SnapBackpack instance that is currently in the level on the Backpack Path slot

3D Model is taken from the Quaternius Survival Pack, it was further edited by DigitalN8m4r3 License: CC0 Public Domain

Note:
the initial pr had a load and onready present in the script for the backpack.
now the hardcoded part was removed to make this more generic.

It would be of course nice if we could make this whole thing even more generic but i fear this might not be that easy


DigitalN8m4r3:
depends on the following PR #401 
the one above needs to be merged before this PR is being merged

PR #398, opened 2023-02-27
flapping improvements: add altitude limit and allow roll-turning while flapping

surreal6:
I added an altitude limit when flapping

Also changed the order of the roll-turning calculation and the falling check that was preventing to turn while gaining altitude after flapping.

### DIFF
--- a/addons/godot-xr-tools/functions/movement_glide.gd
+++ b/addons/godot-xr-tools/functions/movement_glide.gd
@@ -128,7 +128,8 @@ func physics_movement(delta: float, player_body: XRToolsPlayerBody, disabled: bo
 	var allowed_flapping_altitude = false
 	if wings_impulse:
 		if is_sphere_planet:
-			allowed_flapping_altitude = sphere_planet_center.distance_to(player_body.translation) < max_flapping_altitude
+			var altitude = sphere_planet_center.distance_to(player_body.translation)
+			allowed_flapping_altitude = altitude < max_flapping_altitude
 		else:
 			allowed_flapping_altitude = player_body.translation.y < max_flapping_altitude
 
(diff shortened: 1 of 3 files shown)


Malcolmnixon:
This assumes "altitude" can be evaluated by just a Y axis comparison. This wouldn't work for sphere-world or other environments where gravity/down can vary.

A more flexible option would be to define an Area3D where the player is allowed to fly. This would not only allow normal height-limiting, but could also define a flyable sphere around a planet.

surreal6:
ok, it would be nice to have the flapping feature working in the sphere planet, but right now it doesn't work as the arms position also assume Y axis as altitude... 

I will try to change the complete code to make it dependant of the up_gravity_vector, but  the roll-turn improvemente really improves the experience in the flat world scene.

It's ok if remove the altitude check to have the other improvemenet accepted in the meanwhile?

surreal6:
never mind, i found a way to do it with local coordinates, so now it works and no need to check the up_vector.

I was about to export the area as you mention but find easier to just define a gravity center point.

lunarcloud:
I really like the altitude limit idea, but I think there needs to be more guidance docs from the project maintainers on what's allowed / required / wanted. In my opinion, the whole exercise of implementing gravity-switching broaches "project-specific" territory.
I like your point solution more than an area-based one, personally, but it all feels like there need to be a "leave it to the game to implement differently" or "it's done *this* way in our toolkit" contributor guide.

Because what if gravity isn't a sphere, but it's just... LEFT now? What if there's MULTIPLE spheres to walk on?

<hr>

I'd like to see the roll-turn part separated out so it can be pulled in sooner, outside of the altitude-limit discussion.

Malcolmnixon:
As all movement providers have a common "enabled" property, we could add some new nodes that would work for all types of movements such as XRToolsMovementEnableInArea, or XRToolsMovementEnableByHeight. These nodes could be attached as children of any type of XRToolsMovementProvider (flying, gliding, grappling, sprinting, etc.)

Issue #400, opened 2023-02-28
Snap zone should support scripted request to snap an item


Malcolmnixon:
The XRToolsSnapZone script has a function to safely request it pick up an object; however that function is called `_pick_up_object` which is private and gdlint complains of any attempt to call it. The function should be renamed to `pick_up_object` so it can be called externally.

DigitalN8m4r3:
indeed, in my case regarding the basic inventory, this ends up with exactly 2 errors related to this issue and from talkin to malcolm, there is no way around the _pick_up_object, it needs to be made public so that future work such as the backpack inventory and some other things can be done properly

Malcolmnixon:
This works as intended and has been added to the pickable demo scene.

PR #401, opened 2023-03-01
Added return-to-snap-zone script

Malcolmnixon:
This pull request adds an XRToolsReturnToSnapZone script which returns an XRToolsPickable object to a snap-zone if dropped for longer than a specified duration. Additionally this pull request implements feature #400.


DigitalN8m4r3:
lgtm, no errors now with gdlint when returning to snapzone

BastiaanOlij (approved):
Ohh this is cool!

PR #402, opened 2023-03-01
added basic backpack inventory

DigitalN8m4r3:
this adds a slot based inventory backpack with snap require and exclude setup

How To add the inventory to your existing scene:
select the BackpackSnapZone, located inside the xr tools objects/inventory folder, drag it under the Camera, the BackpackSnapZone is already configured to be located around your Neck/Back area, so once you grab behind your head, you will be able to pickup the holstered Backpack.

select the inventory_backpack.tscn that is located inside the objects/inventory folder of the xr tools, drag it under the Origin node (ARVROrigin/XROrigin), next select the snap_backpack.tscn from the objects/inventory folder of the pickable_demo, drag into the scene, now select the InventoryBackpack and drag the previously added snap_backpack under the backpack_path of the InventoryBackpack.

- circle shaoed snapzones are applying scale to the snapped items, this is done thru their parent(s), as they have been scaled.
- oval shaped snapzones are limited to weapons of types heavy/knife


DigitalN8m4r3:
on a side note:
i disabled the beltsnapzones inside the pickable_demo
reason being simply that they behave way to strange, once you move your hand/s forward/backward, they swing along and this is no behaviour we would want inside a beltsnapzone

will file a bug report but seeing that we might soon switch to 4.0 as master, there are obviously more interesting things

BastiaanOlij:
From the videos I've seen so far of this working this is amazing.

However this is an implementation of an item using XR tools components, I'm not sure if that belongs in the tools library itself. Some of the improvements to make this work certainly do, but the backpack itself not. People will want to make their own backpack with their own backpack model, or use the same approach for a duffel bag, or some other storage device and use the backpack as an example of how to put something like this together using XR Tools. 

It's similar as to my own case where I'm using snapzones to implement a reload system for a weapon, or using this to knock an arrow to a bow.

These things belong in the demo and I think we should move it there.

When discussing this with Malcolm he also made the observation it would be a good idea to add these things to the template as well and we should create similar structures so it's easy to add things from the demo to the template in a consistent way (though I think the template should be reorganised first)

DigitalN8m4r3:
alright, will move the inventory folder to the pickables demo. no model was present inside that folder and the backpack scene itself was already part of the pickables demo.

My thought was since this is supposed to be a Toolkit, to add the functionality to it but we clearly do not share the same view here.

DigitalN8m4r3:
closed this pr since i do not intend on updating again after braking changes got merged right before i made the required changes and since the functionality in question is not to be part of the xr tools anyways it makes no sense to do update this again.

For the ones interested, this might return in the form of a template with inventory saving but not as part of the xr tools and its demo scenes

PR #405, opened 2023-03-02
added tool annotation to snap_zone.gd

### LINKED ISSUE #404: snap_zone.gd needs a tool annotation - without it the radius shown is incorrect
issue:
snapzones are shown as being 1 meter in radius, this is incorrect and the reason for that seems to be a rendering artifact

fix:
add the tool annotation to snap_zone.gd

DigitalN8m4r3:
added tool annotation to snap_zone.gd
this fixes the rendering artifact that occurs in editor

without it, the radius of the snapzone is shown to be 1 meter

this PR fixes #404

### DIFF
--- a/addons/godot-xr-tools/objects/snap_zone.gd
+++ b/addons/godot-xr-tools/objects/snap_zone.gd
@@ -1,3 +1,4 @@
+tool
 class_name XRToolsSnapZone
 extends Area
 


Malcolmnixon:
This looks good to me - with this the snap-zones update correctly in the editor.

BastiaanOlij (approved):
May need to check if we need to add a few extra checks so code that shouldn't run in the editor isn't run, but I'm happy for adding tool features to this.

@Malcolmnixon maybe you can have a closer look?

PR #420, opened 2023-03-15
Create character body centric player body implementation

BastiaanOlij:
Creating a new character body centric player body and adapting our movement provider implementation so it works with both old and new approach.

Needs a LOT of cleanup still...


DigitalN8m4r3:
had time to test this, it is promissing on paper but like u stated it needs cleanup and right now it is not useable... at least not for what i would love to use it.

my use case would be for entering/exiting vehicles, once you enter the vehicle i would like to prevent the player from being able to change the height of the camera if the player stands up or walks physicly i want the player to stay at the entering position.

i do have already a system in place to handle most of the things i need but the above mentioned is a dealbreaker, so i hope that the player centric work can be useful to solve the issues i got.

BastiaanOlij:
We're making this the basis of the rebuild XR Tools v2. This will be closed once that goes public.

DigitalN8m4r3:
> We're making this the basis of the rebuild XR Tools v2. This will be closed once that goes public.

when ur saying Basis, that means it wont end up in v1, just to clarify?!

BastiaanOlij:
> > We're making this the basis of the rebuild XR Tools v2. This will be closed once that goes public.
> 
> when ur saying Basis, that means it wont end up in v1, just to clarify?!

Indeed, the problem in adding this to v1 almost means a full rewrite, which is what this PR made a start to, and it is far to disruptive.

BastiaanOlij:
I'm closing this as I wont be returning to it for XR Tools.  The impact is just to breaking and large.

XR Tools 2 is using this approach and I've got several videos on how to scratch build in case people want to have a closer look.

PR #421, opened 2023-03-16
Fix staging fade shader so it works on both Vulkan and GL renderers

Malcolmnixon:
This pull request adds `cull_disabled` to the fade shader so it works with the handedness difference between Vulkan and GL.


Starkium (approved):

BastiaanOlij (approved):
lgtm

PR #422, opened 2023-03-17
Fixed main menu level having to load all other levels

Malcolmnixon:
The teleport object originally used a PackedScene to let the developer specify which scene to teleport to. This was wasteful as it resulted in the main menu having to import all other scenes. This pull request replaces the PackedScene with just a file-path string to the tscn file.

### DIFF
--- a/assets/meshes/teleport/teleport.gd
+++ b/assets/meshes/teleport/teleport.gd
@@ -6,8 +6,8 @@ extends Node3D
 ## Scene base for the current scene
 @export var scene_base: NodePath
 
-## Scene to teleport to, or none for main menu
-@export var scene: PackedScene
+## Main scene file
+@export_file('*.tscn') var scene : String
 
 ## Title texture
 @export var title: Texture2D: set = _set_title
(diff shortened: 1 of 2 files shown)


Starkium:
I checked all this and while it "works", I'm not content with how hard the load stutter is. I think it'd be wise to create some asynchronous thread loading setup here instead.

BastiaanOlij:
> I checked all this and while it "works", I'm not content with how hard the load stutter is. I think it'd be wise to create some asynchronous thread loading setup here instead.

It's supposed to be using asynchronous thread loading but I think that doesn't work well on Godot 4 yet as the new renderer has limited support for pushing data onto the GPU through a separate loading queue.

BastiaanOlij (approved):
Good find Malcolm, indeed using packed scene preloads everything so all our demos are suddenly in memory. Not an ideal situation by a long shot.

I figure this will work well on Godot 3 but have some issues in Godot 4 that need to be fixed upstream.

Malcolmnixon:
Any "large" game cannot load all the scenes in memory at the same time. As such even though this causes stuttering when loading in 4.0, I agree the correct solution is to fix the Godot 4 asynchronous loading.

PR #425, opened 2023-03-26
Add XR Tools editor menu

Malcolmnixon:
This pull request adds an XR Tools menu to the editor with the following settings:
 - Enable OpenXR 
 - Set Physics Layers


Malcolmnixon:
Note: If we agree that the grapple-layer should be supported (https://github.com/GodotVR/godot-xr-tools/pull/426) then this PR should be modified to name that layer as well.

Issue #179, opened 2022-08-16
Grapple hook needs a blocker


Malcolmnixon:
Currently the grapple hook target only collides with layers selected for grappling. The result is that it's possible to shoot "through" an object that isn't grappleable and hit an object behind it.

Instead there should be two masks - one for what objects the grapple-target collides with, and another specifying which objects are valid grapple-targets

PR #426, opened 2023-03-26
Add finer control of grappling hook layers

Malcolmnixon:
This pull request implements feature #179 by:
 - Adding a standard grappling-target physics layer (layer 5)
 - Adding a grapple_enable_mask to the grapple movement provider
 - Modified the demo grappling platform and tower to be on grappling-target layer

With these changes, the grappling hook can be configured to specify what layers the grapple "collides" with, and separately what layers grappling is "enabled" for:
image

An example is that the grapple would collide with the ground, but you would probably not want the player grappling along the ground.


BastiaanOlij (approved):
lgtm

PR #435, opened 2023-04-02
Added basic swim demo from work performed by pietru2004.

Malcolmnixon:
This pull request is a cleaned up version of [Pietru2004's Water Demo](https://github.com/GodotVR/godot-xr-tools/pull/393).

For now it removes any attempts at advanced shading such as Volume Fog, and instead concentrates on the actual swimming movement. Rendering improvements can be added later.


Malcolmnixon:
I'm temporarily converting this back to a draft - a dedicated swimming movement provider is being developed in https://github.com/Malcolmnixon/GodotWildJam-56 which actually uses hand-movements to control swimming.

pietru2004:
_On my defence I am in middle school and sometimes teachers are bit strange._

Anyway it is cool to hear that there will be dedicated swim, personally I want to say that I see both ways of valid - dedicated and using fly movement as swim.

BastiaanOlij:
@Malcolmnixon will have another look at this in the near future. We have a number of swimming implementations that we can revisit and see if we can properly embed into XR Tools.

PR #450, opened 2023-04-12
Add render target size multiplier option to StartXR

Malcolmnixon:
This pull request adds the Render Target Size Multiplier option to godot-xr-tools. This requires Godot 4.0.3 or later (https://github.com/godotengine/godot/pull/73558).

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -2,6 +2,7 @@
 - Enhanced grappling to support collision and target layers
 - Added Godot Editor XR Tools menu for layers and openxr configuration
 - Improved gliding to support roll-turning while flapping
+- Added render_target_size_multiplier to StartXR (requires Godot 4.1+)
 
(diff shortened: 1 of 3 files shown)


BastiaanOlij:
Need to double check if this works with the newly release 4.0.3, if so we should merge this.

Malcolmnixon:
> Need to double check if this works with the newly release 4.0.3, if so we should merge this.

I just checked and it wasn't back-ported into 4.0.3. Looks like we may need to wait for 4.1.

Malcolmnixon:
The render target size multiplier was merged into the 4.1 release (see https://godotengine.github.io/godot-interactive-changelog/#4.1) and I just tested this works on 4.1.1

Malcolmnixon:
We still have many people using Godot 4.0.3, so this now detects whether the feature is implemented.

PR #456, opened 2023-04-19
Cleanup of environments and physics settings

Malcolmnixon:
This pull request:
 - Removes the "default_env.tscn" environment setting from the scene_base.tscn addon
 - Adds the "default_env.tscn" envrironment to all the demo scenes
 - Updates the physics layer names to match those set by the plugin script
 - Updates the physics penetration settings to the new recommended default


BastiaanOlij:
Maybe rename `default_env.tres` to `demo_env.tres` and see if we can use `get_configuration_warning` to alert a user they haven't selected an environment.

BastiaanOlij (approved):

Issue #461, opened 2023-05-04
IDEA - XR seats


pietru2004:
I think it would be cool to have seats logic, it could allow to make XR Vehicles.

PR #462, opened 2023-05-04
Seating_v1

pietru2004:
So I was thinking how to make easily seats in XR.
And here is what I thinked of.

https://github.com/GodotVR/godot-xr-tools/issues/461


DigitalN8m4r3:
didnt tested this yet but i do wonder about one thing, is it restricting the camera/player from changing the height if the user is seated?

pietru2004:
for now all it does is disable kinematic body and snap player to location
I still need to add process check that if seat is moved it will update player accordingly
as for height you in theory can walk out from seat if you have enough space

pietru2004:
Also I try to not statically bind player position - I already seen problem that player can't lean on table or over barrier.

BastiaanOlij:
In XR Tools 2 we are working with a separate dynamic (for walking around) and static (for seated) rig, we need to build a nice solution for transitioning between the two.

Completely basing the static rig on a disabled player body has a number of issues going forward especially because you're likely to have very different controls and options in seated mode v.s. walking mode.

We're currently not thinking of doing anything in XR Tools 1 for this.

PR #463, opened 2023-05-05
Desktop support for XR Tools

pietru2004:
I start this in draft so it won't be yet merged and I can get feedback.

I fought that some game devs might want add support for the desktop mode of their game(If the game was started as XR tittle) and it would be best to keep one character controler rather then have 2 different ones - one for XR and second created by the dev for desktop.

**Note:** At current state only turn and direct are available - others have only files prepared

Also having XR Desktop could allow for faster testing, especialy if your XR headset takes some time to setup or messes with graphic drivers (looking at WMR - for example I can't take screenshoots normal way when having WMR active).


pietru2004:
added jump and sprint - both are temporary placed under the main_menu_level

pietru2004:
erm...

pietru2004:
Quick summary - tools to allow testing of levels without plugin headset, cause sometimes there might be too litle light in room, you might be lazy, or just headset would be unpractical for something to test.
Also it can allow to implement game support for both XR and VR.
Applied successfully in 1 personal project.

BastiaanOlij:
Had a quick look at this during todays XR meeting. This needs more people testing and providing feedback, but generally it's a decent approach to offering both desktop and VR modes for a game.

Should consider this depending on feedback.

surreal6:
Hi! I think this is a good solution to develop a game both in flat and XR modes. 

kudos @pietru2004  and thanks for your work!

I tested direct movement, jump, sprint, crunch, strafe, flight, and function pointer. All of them seems to work fine for me.

Didn't tested gravity zones... not sure how to test it.

I think we could do something similar to @decacis' vr-toggle demo, where the vr-player is replaced by a flat-player at the startup, through code. That way we can change any function with the related desktop function at startup or everytime a scene is loaded in case you use staging.

Maybe in startXR file we can export a variable to select if we want flat-mode fallback or quit game if no xr interface is found.

I have some crash with the rumble_manager, if you enter in a teleport crouching. I assume it's just becouse this is a wip.

I tested this in 4.2.2.stable btw.

pietru2004:
@surreal6 gravity zones is small helper I made for space based game it is like wall walk, but uses Arena3D to get gravity field and if player is not in any gravity it just enables flying for them, if player is in gravity it sets to that gravity dir, Arena3D are grouped into 3 groups of priority, `gravity_zone1` group being smallest priority and `gravity_zone3` highest.

pietru2004:
Also my desktop functions only for when xr is off so player could take off headset and later put it back on...

BastiaanOlij:
I'm pretty happy with how this is put together. My only nitpick is that I don't think checking for our staging node is right. Not everyone will be using that.

We'll need to find a  better way to expose the xr_active flag on our start xr script, maybe through a static method or something. @Malcolmnixon any ideas?

Happy for that to be a follow up PR seeing we've left this too long already.

BastiaanOlij (approved):
Pending linter issues being resolved, and probably could use a rebase.

pietru2004:
@BastiaanOlij I think I fixed gdlint issues... also I added FunctionDesktopPointer to the main_menu_level along with trigger_left action mapping.

DigitalN8m4r3:
> I'm pretty happy with how this is put together. My only nitpick is that I don't think checking for our staging node is right. Not everyone will be using that.
> 
> We'll need to find a better way to expose the xr_active flag on our start xr script, maybe through a static method or something. @Malcolmnixon any ideas?
> 
> Happy for that to be a follow up PR seeing we've left this too long already.

not malcolm but how about the same way we did it with the game state in the xr template?
by using a singleton
[https://github.com/GodotVR/godot-xr-template/blob/13fba1b8ea75b7802660e6627347e9d438dfd537/game/game_state.gd#L4](url)

Malcolmnixon:
I think we'll just get it merged first, and then do a little cleanup around the xr_active and a few grammatical issues.

PR #464, opened 2023-05-15
Change environment handling so we can use WorldEnvironment in levels

BastiaanOlij:
This PR is a simplification of the way we use environments with the added benefit that we see the correct environment when editing levels.

Instead of us using an environment set on our base scene and assigning this to a world environment through code, our level scenes now have their own world environment. This is now possible as we're no longer using the XROrigin/XRCamera setup in staging but expect each level to configure their own player setup.

Our staging scene uses an environment set on the XRCamera that overrule any currently loaded world environment and ensures our loading scene has a black background. This is the only environment not seen in the IDE when editing.

To further simplify things, the holodeck map scene now contains a world environment with the holodeck environment. This ensures that levels using the holodeck automatically use the correct environment.
Similarly our basic map scene now has a world environment with our demo environment set.

The spherical world has it's own world environment and there are one or two other levels that use the demo environment without using the basic map.

This is a minor breaking change for existing games that use XR Tools and rely on the staging system.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,3 +1,6 @@
+# 4.2.0
+- Environments can now be set normally in scenes loaded through the staging system.
+
 # 4.1.0
 - Enhanced grappling to support collision and target layers
 - Added Godot Editor XR Tools menu for layers and openxr configuration
--- a/addons/godot-xr-tools/staging/scene_base.gd
+++ b/addons/godot-xr-tools/staging/scene_base.gd
@@ -30,13 +30,6 @@ signal request_load_scene(p_scene_path)
 signal request_reset_scene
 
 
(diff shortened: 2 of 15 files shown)


Malcolmnixon (approved):
There's one oddity in scene_base.tscn where it added an "environment = null" line even though you removed the environment export from scene_base.gd.

Other than that it looks fine. This will certainly help designing scenes.

BastiaanOlij:
> There's one oddity in scene_base.tscn where it added an "environment = null" line even though you removed the environment export from scene_base.gd.

Yeah it sometimes does that when it hasn't figured out on loaded scenes that a property was removed from a base scene.

Fixed it up

PR #468, opened 2023-05-24
233 - add dead zone to user settings

surreal6:
I added x and y deadzone values and applied in movement direct

also added this attributes to user preferences, and added to the main level UI

I believe it can be improved as now it is checking always to react to the virtual selector

I will add some comments in the code to ask a few questions

### DIFF
--- a/addons/godot-xr-tools/functions/movement_direct.gd
+++ b/addons/godot-xr-tools/functions/movement_direct.gd
@@ -44,7 +44,8 @@ func physics_movement(_delta: float, player_body: XRToolsPlayerBody, _disabled: 
 	var dz_input_action = XRToolsUserSettings.get_adjusted_vector2(_controller, input_action)
 
 	player_body.ground_control_velocity.x += dz_input_action.x * max_speed
-	player_body.ground_control_velocity.y += dz_input_action.y * max_speed
+	if strafe:
+		player_body.ground_control_velocity.y += dz_input_action.y * max_speed
 
 	# Clamp ground control
 	var length := player_body.ground_control_velocity.length()
--- a/addons/godot-xr-tools/functions/movement_direct.gd
+++ b/addons/godot-xr-tools/functions/movement_direct.gd
@@ -28,6 +28,14 @@ extends XRToolsMovementProvider
 # Controller node
 @onready var _controller := XRHelpers.get_xr_controller(self)
 
+## y_axis_dead_zone (from configuration)
+@onready var _y_axis_dead_zone : float = XRTools.get_y_axis_dead_zone()
+
+## x_axis_dead_zone (from configuration)
+@onready var _x_axis_dead_zone : float = XRTools.get_x_axis_dead_zone()
+
+@onready var label = $"../../XRCamera3D/Label3D"
+
 
 # Add support for is_xr_class on XRTools classes
 func is_xr_class(name : String) -> bool:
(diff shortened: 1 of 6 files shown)


addons/godot-xr-tools/functions/movement_direct.gd
```diff
@@ -40,13 +48,34 @@ func physics_movement(_delta: float, player_body: XRToolsPlayerBody, _disabled:
 	if !_controller.get_is_active():
 		return
 
+	# this is not eficient, should receive a message to update or something...
```
surreal6, line 51: how to update this values? with a signal from the UI on the apply?

addons/godot-xr-tools/user_settings/user_settings.gd
```diff
@@ -42,6 +48,8 @@ func _ready():
 func reset_to_defaults() -> void:
 	# Reset to defaults
 	snap_turning = XRTools.get_default_snap_turning()
+	y_axis_dead_zone = 0.1
+	x_axis_dead_zone = 0.2
 	player_height_adjust = 0.0
```
surreal6, line 53: it is posible to access the _define_project_setting/default value to avoid hardcode this?

addons/godot-xr-tools/functions/movement_direct.gd
```diff
@@ -40,13 +48,34 @@ func physics_movement(_delta: float, player_body: XRToolsPlayerBody, _disabled:
...
+	_y_axis_dead_zone = XRTools.get_y_axis_dead_zone()
+	_x_axis_dead_zone = XRTools.get_x_axis_dead_zone()
+
```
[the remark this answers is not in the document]
BastiaanOlij: We could introduce a `deadzone_changed` signal for this in `XRTools.gd`, we then subscribe to this signal in our `_ready`. In the xrtools script we simply emit that signal whenever the deadzone changes.

addons/godot-xr-tools/user_settings/user_settings.gd
BastiaanOlij: Not that I'm aware off..

BastiaanOlij: Actually, looking at this, player adjust is something specific to the user settings.

You need to look at how snap_turning works, the idea here was always that the settings in XRTools form a base set through project settings, and usersetting override this, they aren't means to be written back into the project settings.

It's the usersettings that are then used in the application.  That also removes the requirement of signals because usersettings always contain the values directly accessible.

I think you can even add a helper function `get_adjusted_vector2(p_controller, p_input_action)` to user settings that processes the deadzone and returns an adjusted vector. That way it can be more widely used than purely in the movement script.

surreal6: Ok, now its more clean solution. I was mixing project and user settings concepts. Now it kepts updated 

I kept the declaration of set_y_axis_dead_zone and set_y_axis_dead_zone but they are not used anywhere, the same as other properties

addons/godot-xr-tools/plugin.gd
```diff
@@ -84,6 +84,22 @@ func _enter_tree():
...
+			"0.0,0.5,0.01",
+			0.1)
+
+	# Add input x_axis_dead_zone to the project settings
+	_define_project_setting(
+			"godot_xr_tools/input/x_axis_dead_zone",
+			TYPE_FLOAT,
+			PROPERTY_HINT_RANGE,
+			"0.0,0.5,0.01",
+			0.2)
```
BastiaanOlij, line 101: is it correct our default here is 0.2?

BastiaanOlij (approved):
This looks good.

Note that depending on whether we merge this first, or #475, we could get a nasty merge conflict as I added margin containers to the UI. It may work though..

addons/godot-xr-tools/plugin.gd
surreal6: when testing, I found useful adding a bit more deadzone in the x axis to avoid accidental rotation when moving forward... 

the original issue suggested a value of 0.2.  I'm not sure which value would be more convenient by default

BastiaanOlij: Owh, yeah that makes sense :)

Malcolmnixon:
Unfortunately I just found this PR has a bug in movement_direct.gd

In the original code:
 - Y (forwards backwards) was always applied
 - X (strafing) was only performed if strafing was enabled.
 
In the new code:
 - X (strafing) is always applied
 - Y (forwards backwards) is only performed if strafing is enabled.

I'll file a bug shortly.

PR #469, opened 2023-05-28
Audio demo

DigitalN8m4r3:
This adds my work on audio in xr
it contains new nodes such as the pickableaudio node and the areaaudio node.
Both can be used to play audio
Snap Zone was edited to allow for playing audio upon snap/ stash

pickable audio plays audio based on the velocity of the rigidbody3d

area audio can be attached to all nodes of type area3d or nodes based on the area3d node, it plays audio upon entering
and can be used for things like buttons..
snap zone can play a audio file if the snap zone contains a audiostreamplayer3d node and if not it will pass, maybe someone else could do this with some more elegance

The Demo is showcasing what can be done with the above

Note:
If you have ever experimented with audio for xr, you will know that there is alot that can be done and that can be done wrong
For now i have set all the pickables to Sleeping and it checks if its not sleeping and then it will run the process.

The pickables audiostreamplayer3D nodes are adjusted strictly to specific values, i spent some time gettin these for my ears right so you can of course try it with the standard settings but i do think that you might favor mine

Final Note:
an audio is only as good as the audio file that is being used, keep that in mind


BastiaanOlij:
First impressions is that this looks really good. I really like the demo, it's a good exposition of what you can do with XR tools.

Just a nitpicking question, why did you keep the `AudioStreamPlayer3D` separate? I would have either inherited from `AudioStreamPlayer3D` or created it as a child (though you'd need to duplicate a number of key audio properties)?

For the snapzone stash sound I would just add the audio player to `snap_zone.tscn`, if it's unused it just doesn't contribute to the audio. 

I would also make the 3 sounds in `XRToolsPickableAudioType` optional, I might not want to specify all 3.  I'm also not sure if it's overkill to have this as a resource type instead of just having the sounds be properties of `XRToolsPickableAudio`, I'm not sure if there are a lot of scenarios where you would reuse these. Generally speaking you would add this node to an inherited scene for the pickable object and that already becomes a suitable container. But I might be overlooking some use case.

DigitalN8m4r3:
**_why did you keep the AudioStreamPlayer3D separate?_**
honestly it was because i was thinking about makinga visual Fx node and connect it with the audio node i made but you are actualy right there is no reason to make it seperate, thinking about it am gonna make this extend from the audiostreamplayer3d node

**_For the snapzone stash sound I would just add the audio player to snap_zone.tscn, if it's unused it just doesn't contribute to the audio._**
Hmm am gonna look into this, wasnt sure how to code it actualy since redesigning current code should not end up in a hacky solution, that is why i introduced the check if a audiostreamplayer3d is present do this and if not just pass

**_I would also make the 3 sounds in XRToolsPickableAudioType optional, I might not want to specify all 3. I'm also not sure if it's overkill to have this as a resource type instead of just having the sounds be properties of XRToolsPickableAudio_** 
not sure if i can make this optional :D and regarding the sounds being part of a resource, think about for example all the wooden items u got or that are made out of wood, you just drag in the wood resource and ur done but maybe am not understanding the impact of using the resource type here, so if its making a higher impact when using a resource then am gonna change it

On a sidenote, what about the floor material? am asking cuz am not sure how performant it is so maybe u can take a look at it

surreal6:
i was testing the pr and it works fine for me. I think it's a nice addition to the demos. Also find interesting to have the separate XRToolsPickableAudioType so many objects can share same sounds, and also as an example on how to setup this resources.

DigitalN8m4r3:
thanks to @BastiaanOlij who was able to transfer the arcade floor material into a shader, this now performs way better since the material version i made had quiet a lot going on to achieve the effect.

with that being said, i see this demo complete and ready for review.
however am sure i will revisit this some time later in the future to add a proper demonstration for the footstep stuff since this demo replaces the footstep demo as well

BastiaanOlij (approved):
I think this is good enough to merge, @Malcolmnixon if you could have a look as well and give your Ok?

DigitalN8m4r3:
since malcolm wasnt added to do a review i guess he just didnt rly noticed this, anyways... as he is not available in the coming weeks and since @BastiaanOlij already approved, i would like to see this merged so i can get to the other things such as the collision hands and twohanded stuff

BastiaanOlij:
Lets go for it, I'm sure we can polish up anything that we need later, thanks for the effort @DigitalN8m4r3 !

PR #470, opened 2023-06-07
Fix push bodies

### LINKED ISSUE #467: Lost push rigid body feature
It seems we lost our push rigid body feature in our player body because some of the related functionality was lost in Godot 4s `CharacterBody3D` implementation.

After asking around there is a solution worth trying out. It's detailed in a kidscancode article here: https://kidscancode.org/godot_recipes/3.x/physics/kinematic_to_rigidbody/index.html

Basically we loop through the colliders identified during the execution of `move_and_slide` and apply the needed force. I think we can improve upon this logic.

BastiaanOlij:
Re-implement being able to push bodies around with the player body

Fixes #467

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,5 +1,6 @@
 # 4.2.0
 - Environments can now be set normally in scenes loaded through the staging system.
+- Fixed issue with not being able to push rigid bodies when colliding with them.
 
 # 4.1.0
 - Enhanced grappling to support collision and target layers
--- a/addons/godot-xr-tools/player/player_body.gd
+++ b/addons/godot-xr-tools/player/player_body.gd
@@ -45,6 +45,7 @@ const NEAR_GROUND_DISTANCE := 1.0
 ## If true, the player body performs physics processing and movement
 @export var enabled : bool = true: set = set_enabled
 
+@export_group("Player setup")
 ## Radius of the player body collider
 @export var player_radius : float = 0.2: set = set_player_radius
 
@@ -63,8 +64,15 @@ const NEAR_GROUND_DISTANCE := 1.0
 ## Mix factor for body orientation
 @export_range(0.0, 1.0) var body_forward_mix : float = 0.75
 
+@export_group("Collisions")
+
 ## Lets the player push rigid bodies
 @export var push_rigid_bodies : bool = true
+
+## If push_rigid_bodies is enabled, provides a strength factor for the impulse
+@export var push_strength_factor : float = 1.0
+
+@export_group("Physics")
 
 ## Default ground physics settings
 @export var physics : XRToolsGroundPhysicsSettings: set = set_physics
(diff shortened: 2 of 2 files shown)


Malcolmnixon:
There are a few problems with the impulse application logic.

## Impulse Strength
Currently the impulse strength doesn't vary based on whether the player is walking into the object, or sliding almost perpendicular to the object. This one's a fairly simple fix - just change the strength calculation to be based on the dot-product between the player velocity and the surface normal as follows:
```gdscript
var strength = p_velocity.dot(-with.get_normal(0)) * push_strength_factor
```
This feels more natural when walking into objects in glancing blows vs straight on.

## Low Mass Bullets
Another problem which is more challenging is that if you try pushing a RigidBody3D with a low mass (0.0001) then it will fire off faster than a bullet, as the "impulse" results in a massive velocity. You want heavier objects to be more resistant to moving, but you don't want lighter objects to be affected so much that they move faster than the player. I'm not sure how to fix that problem.

## Relative-Velocity Impacts
A third problem is with reference-frames. If the player is in a train moving at 10m/s and just brushes up against an object, the p_velocity of the player will be around 10m/s, even though the velocity "at the point of contact" is tiny. The solution is to change from using the players absolute velocity, to instead use the relative velocity between the player and the RigidBody3D contact point. The problem is it seems to be difficult to ask Godot what the contact point velocity is. There's ```PhysicsDirectBodyState.get_contact_collider_velocity_at_position``` which looks roughly correct. It's just odd that KinematicCollision3D doesn't expose it.

BastiaanOlij:
@Malcolmnixon sorry for the late response.

Indeed tricky issues, I've adjusted the logic somewhat.

Impulse Strength
===========
The impulse strength change you suggested makes sense so I applied that as you suggested.

Low mass bullets
===========
The issue with low mass bullets is much more tricky. We're talking about transfer of energy as we're hitting an object. So our mass (which doesn't exist in CharacterBody3D) times the velocity of impact determines how much force we're impacting the object with. The lighter the object, the more the resulting velocity is of this energy being applied to the object.

However more is going on than that simplistic highschool view of physics as generally bumping into small objects in your living room, doesn't sent them flying through the room like crazy. I'm not to read up on this so I'm also not sure what the missing ingredient is and how to properly calculate this transfer of energy.

I might see if I can fine some time looking into Godots physics engine because this is no different from what the internal code should be doing when two rigidbodies collide.

Relative-Velocity Impacts
================
The relative velocity thing is an issue, `get_contact_collider_velocity_at_position` seems to be the correct solution but if it's not exposed we'll need to fix that upstream first so we can't rely on it yet. For now I have added that we subtract the velocity of the object we're hitting so we are working with the relative velocity between the two objects. Won't solve the issue for when we're on a moving platform I wager.

Malcolmnixon (approved):
This works very nicely now.

PR #471, opened 2023-06-22
Collision Hand and Two Handed - Demo

DigitalN8m4r3:
**_About:_**
this currently is a barebones setup, that will hopfefuly end up in a demo scene to showcase how to use the above
currently it is not useable in production as it has some issues that would prevent it from being used properly

Anyways here is a quick rundown on how to add the collision hands, as right now there is no demo scene.

**_HowTo:_**
add the collision hand two times under ur xr origing and select the nodes depending on left and right controllers
grafik

if u want to have collision on pickable objects as well, check the pickable_collision mark on the collision hand
and instantiate a PickableCollision.tscn to the pickable object like so

grafik

**_Known Bugs:_**
1) Pickable_Collision shapes are off
2) TwoHanded is getting dropped too fast, needs a way to make this not happen that fast or disable when held with two hands?...
3) if holding a pickable with each hands and those get in contact with each other, they slow down the whole session... will need some further investigation

### DIFF
--- a/addons/godot-xr-tools/objects/collision_hand/collision_hand.gd
+++ b/addons/godot-xr-tools/objects/collision_hand/collision_hand.gd
@@ -1,0 +1,213 @@
+@tool
+class_name XRToolsCollisionHand
+extends CharacterBody3D
+
+#
+# adds collision capabilities to open_xr hands and held objects
+#
+# REQUIRES SETUP:
+# 1) Parent must be ARVROrigin (NOT THE ARVRControllers!)
+# 2) All export NodePaths must be set
+# 3) PhysicsHand and PickupFunction are expected to be children of the controller
+# 4) Held objects may have a get_collider_dict() function to inform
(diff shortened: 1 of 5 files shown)


DigitalN8m4r3:
@BastiaanOlij and @Malcolmnixon 
if you guys should have some time to look at the code, this would be nice

there are some issues and am sure i could use some help on refining some of the things, like the collision shape being offset on the pickable collision of the pickables when using the collision hand but right now am just gonna mark this as draft as i got no demo scene up but am workin on it

DigitalN8m4r3:
dont know what it is that happend but for whatever reason when i try to use pickable_collision and or two_handed with the collision_hand, i get a strange result of the visual hands being totally of in transforms like 5 Meters away

tried without the pickable_collision, using only the collision_hands and the issue didnt happen but once i added the pickable_collision, the whole repo got the issue, i suspect some kind of overshadowing of files in my git stashes.. no idea but i gotta close this pr as it doesnt make sense this way

PR #475, opened 2023-07-02
Refactor player height adjustment

BastiaanOlij:
This PR redoes the player height implementation.

Our default setting in project settings remains however our user settings now store a user overridden version of this.

As our loading screen has the "trigger to continue" logic we can assume the player is standing (or sitting) in "rest" and we calculate the offset required to make the virtual player the height stored in our settings.

This means XR tools can now be played while seated but controlling the character as if standing up. 

Note that this is different from @pietru2004 PRs where we actually want the character to be seated (in a vehicle or otherwise) if I understand correctly. This is something we do want to support properly but I haven't had time to look into this properly.

### DIFF
--- a/addons/godot-xr-tools/player/player_body.gd
+++ b/addons/godot-xr-tools/player/player_body.gd
@@ -46,6 +46,10 @@ const NEAR_GROUND_DISTANCE := 1.0
 @export var enabled : bool = true: set = set_enabled
 
 @export_group("Player setup")
+
+## Automatically calibrate player body on next frame
+@export var player_calibrate_height : bool = true
+
 ## Radius of the player body collider
 @export var player_radius : float = 0.2: set = set_player_radius
 
@@ -56,7 +60,7 @@ const NEAR_GROUND_DISTANCE := 1.0
 @export var player_height_min : float = 0.6
 
 ## Maximum player height
-@export var player_height_max : float = 2.2
+@export var player_height_max : float = 2.5
 
 ## Eyes forward offset from center of body in player_radius units
 @export_range(0.0, 1.0) var eye_forward_offset : float = 0.5
@@ -199,6 +203,7 @@ func _ready():
 	# Propagate defaults
 	_update_enabled()
 	_update_player_radius()
+
 
 func set_enabled(new_value) -> void:
 	enabled = new_value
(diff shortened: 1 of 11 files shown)


pietru2004:
My implementation of seat(newest one) disables playerbody and every frame sets transform origin to seat point+ adds rotation, all on xr origin.

pietru2004:
Also what effect would it have on https://github.com/GodotVR/godot-xr-tools/pull/463 ?

DigitalN8m4r3:
hmm, gotta check this out then
it sounds like what am after, a way to force seated even if the player is actualy standing while preventing from changing height if the player actualy sits down/stnads up while using this. so correct me if am getting this the wrong way

BastiaanOlij:
@pietru2004 sorry I still need to look into your seated solution. It's indeed a bit of a conundrum, player body doesn't work for the situation where the game character is standing up so disabling it is the correct way forward, or possibly this should be a mode.
This PR thus has no impact on it because this PR is to deal with the end user sitting down or being of a different height while we fix the game characters height when standing.

@DigitalN8m4r3 this PR is the opposite, this is the scenario where you're game character is standing up regardless of the end user (player) being seated or standing.

When you look at the scenario where the game character is seated, the game dynamics change drastically from what playerbody is designed for. Controller inputs now no longer effect the game character, say  your game character is seated in a car, controller inputs will effect the car, it's the car that now moves, and the game character as a child node moves with the car.

The problem is anchoring your game character so they are seated in the correct position at the start.

For a game that only allows a seated position this is easy, you place the XROrigin where you want the players head to be and you wait until tracking information is stable and call `XRServer.center_on_hmd(XRServer.RESET_BUT_KEEP_TILT ,false)`. Then you add a button hold feature or something to be able to reset that view by calling this again in case your player wasn't ready.

For a game where you want to be able to switch between a seated and standing position, say you want to make a VR version of GTA, you'll need to implement your own version of what `center_on_hmd` does behind the scenes, When standing playerbody is active and you control the character as we're used to, when the game character enters a car you disable playerbody, child XROrigin to the seat, and change its transform so the head ends up in the correct starting position.

The mistake I see many people make is to try and keep the head in place once the game character is seated and basically turn the game into a 3DOF game. You really really don't want this. When I'm in a car, I want to be able to move my head and look around, this also means moving your head positionally. If you disable this, not only do you induce motion sickness, you also hinder the player from making use of their natural ability to change the angle at which they are looking at something.

This however opens the door to the player moving beyond the space that they can naturally occupy. While hilarious the player can physically stand up and walk away, moving the players head outside of the vehicle.

The best way to handle this situation IMHO is by checking if the head moves out of the acceptable space and fade the screen to black (maybe with a message to press the reset button).
The approach I took in my vehicle tutorial was to simply calculate the distance between the ideal head position and the actual head position and fade the screen to black if this goes beyond an acceptable distance. When faded to black we still show a message to hold the button to recenter the screen.
This could be improved upon by using an Area3D node and a collision shape that denotes where the players head can be, and simply checking if the head is inside of the area.

pietru2004:
What about crouching in real life, mayby game does not use crouch function...

Malcolmnixon (commented):
This looks good - just need to resolve the merge conflicts.

Malcolmnixon (approved):

PR #476, opened 2023-07-03
add render_priority and no_depth_test as option to viewport2d_in_3d

surreal6:
Added options to set no_depth_test and render priority to the mesh material inside viewport2d_in_3d

As the material is created in runtime, we need to expose this values in the script, otherwise changes in the material will not affect the created viewport.

I keep default values as in previous versions.

no_depth_viewport_inspector

no_depth_viewport_use_case

### DIFF
--- a/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
+++ b/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
@@ -47,6 +47,12 @@ const DEFAULT_LAYER := 0b0000_0000_0001_0000_0000_0000_0000_0001
 ## Alpha Scissor Threshold property
 @export var alpha_scissor_threshold : float = 0.25: set = set_alpha_scissor_threshold
 
+## Render priority property
+@export var render_priority : int = 0
+
+## no depth test property
+@export var no_depth_test : bool = false
+
 ## Unshaded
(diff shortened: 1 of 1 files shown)


Malcolmnixon (approved):
This looks good to me. The few times I've played with "no depth test" in VR it's always resulted in interesting effects for example being unable to see hands while in front of the UI; but the developer should have the freedom to enable this if desired.

A alternate approach you might want to consider would be to allow the user to optionally specify their own 'template' material. If not specified the viewport_2d_in_3d would make a new one as it does today, but if the user provides a template tweaked with custom settings, the viewport_2d_in_3d would duplicate it and then override the properties it needs to make the viewport work.

This would give the developer complete control over essentially all material properties - not just the few we've got and the depth/priority you've added.

surreal6:
It's a good idea, i added this as a new enhacement idea in the issues:   #488

BastiaanOlij:
I agree with Malcolm here, we should redo the implementation and expose the material itself, because we're just going to grow settings more and more. 

The challenge will be to let users know we will override some parts of the material (like setting the albedo texture to the actual viewport).

Malcolmnixon:
@surreal6 Could you check out pull request #493 and confirm it meets your needs.

Malcolmnixon:
Closing this on request from @surreal6 now that #493 has been merged.

PR #478, opened 2023-07-04
added Collision Hands Demo

### LINKED ISSUE #136: Feature request: Hand Collision
I think that the player's hands should be able to collide with the environment and objects.

### LINKED ISSUE #158: object weight / drag?
Is there a way to have objects drag behind when picking them up or moving them to simulate their weight?
If not, is this a planned feature?

DigitalN8m4r3:
https://github.com/GodotVR/godot-xr-tools/assets/56046022/ba0dc1ab-25bc-4003-81b7-ec9a3b0b8883

https://github.com/GodotVR/godot-xr-tools/assets/56046022/23e3a3f6-17be-4c19-87f6-f1cd01e7ef67

https://github.com/GodotVR/godot-xr-tools/assets/56046022/feaa6e0d-b221-407e-8dae-614e4b2d785e

adds collision to the Open_XR_Tools_Hands by using a CharacterBody3D that follows the corresponding XRController3D.

it is able to collide with the world and with static objects

optional:
*	collision on pickables
*	weight on pickables
*	zero_g movement

currently not supported:
* accounting for switching grab points

this pull request fixes/closes the following issues/requests:
Closes #158 
Closes #136 

this scene contains a two handed example
and the collision hands are made to account for this, currently the two handed is mainly useable for ranged weapon objects.

however under its additional settings, there is a optional mod. if enabled, changing the rotational offset may produce something along the lines of a melee two handed object.

Additional Note and Thanks - Collision Hands:
this work was based on the CollisionHands work by https://github.com/n8-ball
from the Godot Dojo - https://github.com/Godot-Dojo/Godot-XR-Collision-Hands
Repository.

Additional Note and Thanks - Two Handed:
in the process of makin this work and expand for Godot 4, i received valuable help and input by
https://github.com/stumpynub who explained to me how to get two handed started
@teddybear082 for makin the two handed working
last but not least @Malcolmnixon who as always helped me debug and get the two handed actualy to this state

Finally i would like to extend a welcome to evryone who is interested in this to participate in testing this pr and or adding ideas/ functionality by improving the code

Thanks


DigitalN8m4r3:
feel like am almost done here, might need to change some things here n there but this should be good for a review, the way it is right now.

addons/godot-xr-tools/objects/collision_hand/collision_hand.gd
```diff
@@ -0,0 +1,231 @@
...
+@export_group("Collision Hand Setup")
+
+## select controller
+@export var controller_path : NodePath
+
+## select visual hand
+@export var _hand_path : NodePath
+
+## select function_pickup
+@export var _pickup_path : NodePath #Should be child of ARVRController
```
JD-The-65th, line 29: I would rework these to get the nodes themselves, not the path to the node. It allows for class limiting of the node so that the user can't put in an incorrect node (It also allows for highlighting when selecting a node directly.)

addons/godot-xr-tools/objects/collision_hand/collision_hand.gd
```diff
@@ -0,0 +1,231 @@
...
+	start_speed = _speed
+
+	hand_remote.set_transform(_hand.get_transform())
+	hand_remote.remote_path = _hand.get_path()
+	pickup_remote.set_transform(_pickup.get_transform())
+	pickup_remote.remote_path = _pickup.get_path()
+
+	if pickable_collision:
+		_pickup.has_picked_up.connect(add_colliders)
+		_pickup.has_dropped.connect(remove_colliders)
```
JD-The-65th, line 111: Same here.

JD-The-65th (changes requested):
Hey, your Collision Hands work great! I just have a few concerns before this can be merged.

addons/godot-xr-tools/objects/collision_hand/collision_hand.gd
JD-The-65th, line 29: I would rework these to get the nodes themselves, not the path to the node. It allows for class limiting of the node so that the user can't put in an incorrect node (It also allows for highlighting when selecting a node directly.)

JD-The-65th (changes requested):
Not exactly used to using github reviews, bear with me here :p

JD-The-65th:
One issue I didn't note is that when using the physics hands, the collision doesn't match the fingers, it only matches the fist pose. Aside from that, this works great! No performance hits when using them.

DigitalN8m4r3:
> Hey, your Collision Hands work great! I just have a few concerns before this can be merged.

_First, thank you for taking the time to test and write your feel and concerns about it!_


indeed but this is supposed to be the developers responsibility when using the col hands, however am gonna look up at making the setup automatic

regarding the other thing,
the col hands are made to be generic.
you can as a developer choose to use pickable_collision with or without weight

with that being said, if you do not wish to use pickable_collision, just dont tick the box and it wont add colliders.
but to be able to add colliders in the first place it looks when it picks up a item, if it even has a pickable_collision node and if it got none, it just wont add the colliders

the thing regarding no collision on fingers, if you look at the col hands, they contain a collisionshape3d node that is called palm_shape, this one is soley being used for collision for the hand

the system is limited to this since it would involve a lot more to get collision going for fingers, i do have a idea or 2 on how i would approach this but the way it is now, it is intended to be and the physics hands are not supposed to be used with the col hands.

**_best practice for making collision shapes for pickable items is to make them inside the editor, create a  collisionshape3d node on your pickable or use the exsisting one._**

> select either:

- BoxShape3D
- SphereShape3D
- CapsuleShape3D
- CylinderShape3D

_(of course you can add multiple of these to your pickable, only the first one will be taken into account for collision when being held as the system copys the first collisionshape3d node into a dictionary into the col hand)_

with that being said, the system works just best with simplistic shapes.

surreal6:
First of all, thanks for your work. It 's a lot of effort put inside this pr

I was testing the scene and, while the basic collision works fine, and the pickable collision is also working, it is buggy with the weight option. If you try to raise the 5kg weight it sticks to the table and then jumps to the hand.. it also make the two handed object to fly crazy sometimes and the basket ball also goes crazy if you shake your hand fast.

Also the weight makes the two hand to stick to the floor or whatever object it collides, and then fly around.

Also, if you activate optional weight in right hand and not activated in left hand, then you pick the 1kg witth left hand and 5kg with right hand, you put your arms up and the right hand is always a bit shorter than the other (but it should reach the same height, only in a longer time)

On UX: don't really get why you can't punch the objects with the hand but you can with the gloves, it's a bit confusing. Also, its annoying that you can't go inside the square with the gloves... maybe you can put a second pair of gloves inside the square.

I would recomend to separate the pr into pieces, and merge stable features first, and then put the optional ones in a next step, otherwise it seems to be a big effort to make work everyhing.

surreal6:
btw: i tested it in godot 4.1, but the project is in g 4.0

pietru2004:
for some reason the return portal in collision hands demo did not work, also room is too small, after spawning in it I walked straight into return portal(and it probably bricked it)
it is hard to hit stuff cause player bodydoes not allow to lean over objects

also please move weights to somewhere more visible, I missed where they were at first time
obraz


and It is first time for me seeing new visuals - the holodeckMap is nice... 😄 

and for last thing the: I think the TwoHandedDemoPickable should have bigger snap distance for middle handle, when I move my hand and try to get it from T straight from my body to = parrel to my body it just dismounts if I am not careful enough

DigitalN8m4r3:
> First of all, thanks for your work. It 's a lot of effort put inside this pr
> 
> I was testing the scene and, while the basic collision works fine, and the pickable collision is also working, it is buggy with the weight option. If you try to raise the 5kg weight it sticks to the table and then jumps to the hand.. it also make the two handed object to fly crazy sometimes and the basket ball also goes crazy if you shake your hand fast.
> 
> Also the weight makes the two hand to stick to the floor or whatever object it collides, and then fly around.
> 
> Also, if you activate optional weight in right hand and not activated in left hand, then you pick the 1kg witth left hand and 5kg with right hand, you put your arms up and the right hand is always a bit shorter than the other (but it should reach the same height, only in a longer time)
> 
> On UX: don't really get why you can't punch the objects with the hand but you can with the gloves, it's a bit confusing. Also, its annoying that you can't go inside the square with the gloves... maybe you can put a second pair of gloves inside the square.
> 
> I would recomend to separate the pr into pieces, and merge stable features first, and then put the optional ones in a next step, otherwise it seems to be a big effort to make work everyhing.

i do already have the gloves in the ring for the next commit but am investigating if i can add em as snapables to some kind wristholster (not working currently)

currently the weight for example is a cheap trick that uses apply floor snap and it makes use of the zero g part of the col hands to enhance this effect, it might be cool if we could get it to be raised to where the hand is after some time or after a certain height is reached but i wasnt able to do this yet, still i would want to keep this in the pr as it might just be a metter of time till we can get this working, granted it is not working perfect but i hope to get some help from mux here n there to fix some of the issues

DigitalN8m4r3:
> for some reason the return portal in collision hands demo did not work, also room is too small, after spawning in it I walked straight into return portal(and it probably bricked it) it is hard to hit stuff cause player bodydoes not allow to lean over objects
> 
> also please move weights to somewhere more visible, I missed where they were at first time obraz
> 
> and It is first time for me seeing new visuals - the holodeckMap is nice... 😄
> 
> and for last thing the: I think the TwoHandedDemoPickable should have bigger snap distance for middle handle, when I move my hand and try to get it from T straight from my body to = parrel to my body it just dismounts if I am not careful enough

u got me here, i had to remove the player layer from the playerbody, this is the reason why the return to main menu istn working but therefore other things are working, like for example u can pull urself up and down with the ropes, u can move ur head between two ropes. by using climbables in ur level geometry, u can make a boneworks style of game^^... granted it isnt working perfect yet but it is damn close :D

regarding the twohanded, currently the only reason it is in the pr is to have it for future reference. sadly currently the twohanded breaks its collision on the second hand due to the interactable handle swaying to far away upon collision, even setting a higher snap distance wont change that. however ur more then welcome to look it up and see if u can fix that, i tried till i realized that i dont know how to fix the  handle swaying away when holding it twohanded cuz it isnt doin that with one hand, neither upon collision

BastiaanOlij:
Ok, finally had a good look at this. This is very impressive work, there are some issues that need polishing up but it's a solid base to work on.

One small nitpick is that while I appreciate the artistic license, seeing our target audience is of all ages I think the blood on the mat is a bit much :) I think it's cool, but we do represent the Godot Project so :)

I did do a PR with some structural changes, Digital already merged it, there are a few things to still solve here, so just copying this from my PR:
- The two handed code doesn't look right and I haven't had time to make that work
- The distance check that drops whatever is held is overoptimistic, you can drop the glove by punching the bag and the resulting collision overflow results in the player being thrown to the other end of the room :)
- When the player moves through controller input, the hand isn't repositioned right away because its no longer linked. This is a problem the poke thingy has too.

I'll find some time to keep helping out and working on those issues. 

Note that the change is that `XRToolsCollisionHand` is now a child of the `XRController3D` node, and all the functions, meshes, and such are now children of the collision hand (all these are now designed to work either directly under the controller or under a subnote, with thanks to Malcolms excellent improvements a little while ago).

It still works the same, but the magic is in setting `top_level` to true which means that `XRToolsCollisionHand` no longer gets its position from its parent.

addons/godot-xr-tools/interactables/interactable_handle.gd
```diff
@@ -22,6 +22,20 @@ extends XRToolsPickable
 ## Distance from the handle origin to auto-snap the grab
 @export var snap_distance : float = 0.3
 
+@export_group("Optional attach Hand")
+## this latches the grabbing Hand onto the Interactable
```
Malcolmnixon, line 26: This block of '##' (API Documentation Comments) produces some really funky online documentation.
image

If this is just documentation for developers, you may want to use the '#' style of comments.

DigitalN8m4r3: thanks for taking a look at this, wasnt aware that this would actualy produce this kind of mess.
gonna look into it and remove unecassary comments.

DigitalN8m4r3:
this pr will stay open, simply to showcase another approach to collision hands
am gonna revert back some of the changes that were done to ensure that this works as it did previously as it tested it over the course of weeks with the godot physics engine and with the jolt physics addon, since both performed the same for me.

DigitalN8m4r3:
Closed
no updates to be expected
further work is being done here https://github.com/GodotVR/godot-xr-tools/pull/495

PR #482, opened 2023-07-22
Restore correct movement - forwards/backwards on & strafing optional

Malcolmnixon:
This pull request fixes a bug introduced with the #233 pull request for dead zones.


BastiaanOlij (approved):
Oops :)

Issue #477, opened 2023-07-03
playerbody's stop_on_slope seems not to  work properly


surreal6:
# The characterBody option stop_on_slope seems to fail.

There is no difference if you set this option to true or false, when you move (with movement direct) to a slope, the playerbody is pushed up to the slope. The more sloped, the fastest you move.

## How to reproduce the bug:

Run the example, move to the first slope (yellow one is 5 degrees) and stop moving, you will be moving up no matter if you set stop_on_slope on or off.  

The only way ti found to fix it is to add a custom physics behaviour and uncheck the stop_on_slope in the custom XRToolsGroundPysics

Maybe i am not understanding how to use the option, but it seems to me thats not the intended behaviour

## demo project 

I put together a sample project where you can replicate the bug:

(note that you will need to install xr tools from the asset manager in the demo project)

[https://github.com/surreal6/player_slide_on_slope_bug](https://github.com/surreal6/player_slide_on_slope_bug)

![Captura3](https://github.com/GodotVR/godot-xr-tools/assets/437199/7d7eb569-8a96-4005-80ca-6f0e83379f16)

Malcolmnixon:
This should be fixed now in the main branch. Could you try it out and confirm.

BastiaanOlij:
Closing this as fixed. If there are further issues please open up a new issue with a demo project showing the issue.

PR #483, opened 2023-07-23
Fix slope movement

Malcolmnixon:
This pull request fixes issue #477 and restores normal character movement on slopes.

The CharacterBody3D will not support sliding down slopes due to gravity regardless of the options. What it will do is support sliding on a flat surface. The fix is simply to switch the CharacterBody3D 'up_direction' from the local gravity 'up' to the surface-normal of the ground. This translates the unsupported "sliding down slopes" into "sliding on a flat surface".

image

Diagram 1 - Shows how we were trying to use CharacterBody3D which does not work.

Diagram 2 - Shows how this PR changes CharacterBody3D and makes it work.

Diagram 3 - Same as the second with the observer reference frame rotated showing how the movement actually works - like pushing a piece of paper across a table.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,7 @@
 # 4.2.0
 - Environments can now be set normally in scenes loaded through the staging system.
 - Fixed issue with not being able to push rigid bodies when colliding with them.
+- Fixed player movement on slopes.
 
 # 4.1.0
 - Enhanced grappling to support collision and target layers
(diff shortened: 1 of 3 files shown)


Malcolmnixon:
Bad PR - accidentally had some cruft in there from a different PR.

Malcolmnixon:
Corrected PR is https://github.com/GodotVR/godot-xr-tools/pull/484

PR #484, opened 2023-07-23
Fix slope movement

Malcolmnixon:
This pull request fixes issue #477 and restores normal character movement on slopes.

The CharacterBody3D will not support sliding down slopes due to gravity regardless of the options. What it will do is support sliding on a flat surface. The fix is simply to switch the CharacterBody3D 'up_direction' from the local gravity 'up' to the surface-normal of the ground. This translates the unsupported "sliding down slopes" into "sliding on a flat surface".

image

Diagram 1 - Shows how we were trying to use CharacterBody3D which does not work.

Diagram 2 - Shows how this PR changes CharacterBody3D and makes it work.

Diagram 3 - Same as the second with the observer reference frame rotated showing how the movement actually works - like pushing a piece of paper across a table.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,7 @@
 # 4.2.0
 - Environments can now be set normally in scenes loaded through the staging system.
 - Fixed issue with not being able to push rigid bodies when colliding with them.
+- Fixed player movement on slopes.
 
 # 4.1.0
 - Enhanced grappling to support collision and target layers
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
I don't like it... There has to be a better way to do this. This opens up the door to walking on walls, which in itself is a nice mechanic if you want it but otherwise doesn't make sense. 

Gravity is down, the slide logic in Godot is just.... ugh

BastiaanOlij (approved):
After talking to Malcolm, this seems like an OK workaround. My concerns are mostly addressed in the player body handling code.

Issue #486, opened 2023-07-25
Origin Gravity Demo paths broken on Godot 4.1


Malcolmnixon:
Loading the Origin Gravity Demo scene in Godot 4.0.3 shows the following:
![image](https://github.com/GodotVR/godot-xr-tools/assets/1863707/23d57c46-49fa-4fd2-b471-29d3df0fc426)
Opening the same scene in Godot 4.1.1 shows the following:
![image](https://github.com/GodotVR/godot-xr-tools/assets/1863707/cdc93c79-15ff-4eec-8fbe-e0deae95689b)

The first path is rotated 90 degrees which makes it virtually impossible to test wall-walking.

Malcolmnixon:
Paths now functional in 4.0 and 4.1

PR #487, opened 2023-07-25
Tweak the demo path so it calculates the same for Godot 4.0 and 4.1

Malcolmnixon:
This PR fixes issue #486 by slightly tweaking the Origin Gravity Demo scene so the first walkway builds the same in Godot 4.0 and 4.1.

This is the scene loaded in Godot 4.0.3:
image

And this is the same scene loaded in Godot 4.1.1:
image

The paths aren't perfectly identical, but close enough that they're usable in both versions.


BastiaanOlij (approved):
lgtm

Issue #328, opened 2023-01-11
Poke feature lags behind the finger tip by one physics frame


goatchurchprime:
The poke feature works pretty well, but the way that it lags behind the fingertip when you move your hand really bugs me.  This might be too hard to fix it for now, but I'm logging this as an issue so anyone else who is bothered by it knows what fix attempts won't work.

First I tried using a PinJoint connected between a static object in Poke and the PokeBody.  This should be the way it is implemented, but the Godot physics for PinJoints is terrible in that it works like a sloppy rubber band even in free space.  It seems an `_integrate_forces()` function is defined in PokeBody to implement what the PinJoint should have been doing.

Then I tried a `move_and_collide()` in a `_physics_process()`, and the the poke body was still laggy.  (This should work because it moves it to the position if there is no collision.) Even setting `global_transform.origin = get_parent().global_transform.origin` in a `PokeBody._physics_process()` is laggy.  In fact the PokeBody is only coincident when you set the transform.origin in the `_process()` function.

Obviously, when you set the transform every other process frame (or every three frames) you get a double (or triple) image for the PokeBody, once on the fingertip and once one frame behind.  (It's not possible to screenshot this effect for illustration.)

```GDScript
var n = 0
func _process(delta):
	n += 1
	if (n%2) == 0:
		global_transform.origin = get_parent().global_transform.origin
```

Here's what's really strange: The PhysicsFrame runs at 1/3 the rate of the image frames (20Hz vs 60Hz).  So why don't we get a triple image of the PokeBody when we set the transform.origin in the _physics_frame()?

Aside from this mystery, the only way I can imagine fixing this issue is by faking it: Set the PokeBody/MeshInstance position in the _process() function unless a collision has been detected.

Malcolmnixon:
I've been considering an equivalent workaround for the physics hands when we get to the. - for the same lag reasons

BastiaanOlij:
This should work correctly in Godot 4.

The problem in Godot 3 is that `_physics_process` runs before `_process` and it is within `_process` where we (re)position the controller node. Even in `_process` it isn't until the process of the ARVRController node runs that it is updated in which case it's possible the physics hands or poke nodes `_process` has already been run.

In Godot 4 repositioning is handled before physics runs (unless if you run physics in a separate process in which case nothing will save you).

[...]

PR #489, opened 2023-07-28
Fix the lag in the finger-poke

Malcolmnixon:
This pull request fixes issue #328 by implementing the following changes:
 - Change the PokeBody from being a RigidBody3D driven by `_integrate_forces` to a StaticBody3D driven by `_physics_process`
 - Set the PlayerBody `process_priority` to -100 so its physics processing is performed before anything else

The problem with the old implementation is that `_integrate_forces` is not actually performed in the physics processing phase, and so ignores any priority rules. This means that we cannot enforce the body to move first, and then ensure the poke updates in the same frame with the updated body.

Unfortunately with Godot 4 removing KinematicBody, and CharacterBody3D being too tied to character movement; the move_and_slide functionality had to be implemented in gdscript.

This implementation uses StaticBody3D rather than directly using the PhysicsDirectSpaceState3D/PhysicsServer3D because only PhysicsBody3D objects get the logic to "de-collide" objects in C++, and implementing the loops in gdscript would be crippling.

### DIFF
--- a/addons/godot-xr-tools/player/player_body.tscn
+++ b/addons/godot-xr-tools/player/player_body.tscn
@@ -3,6 +3,7 @@
 [ext_resource type="Script" path="res://addons/godot-xr-tools/player/player_body.gd" id="1"]
 
 [node name="PlayerBody" type="CharacterBody3D" groups=["player_body"]]
+process_priority = -100
 top_level = true
 collision_layer = 524288
 collision_mask = 1023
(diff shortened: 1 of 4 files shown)


BastiaanOlij (approved):
lgtm! probably should apply similar changes to CollisionHand in Digitals PR.

DigitalN8m4r3:
> lgtm! probably should apply similar changes to CollisionHand in Digitals PR.

just tried just setting process to -100 and sadly i got no change there

Malcolmnixon:
The XRToolsPoke node now:
 - gives control over whether it pushes bodies
 - has documentation comments
 - organizes its properties into categories.

image

PR #490, opened 2023-07-30
Implement basic collision hands

Malcolmnixon:
This PR implements the following:
- Extracted force-pushing-body logic from PokeBody into XRToolsForceBody.
- Modified XRToolsHand so it can be any descendant of an XRController3D (from DigitalN8m4r3/BastiaanOlij)
- Added rough XRToolsCollisionHand script
- Modified interactables demo to hack XRToolsCollisionHand in for now

The interactable demo scene puts an XRToolsCollisionHand under the left controller, and moves everything else under it. It also adds a basic box as a hand collider:
image
image

The XRToolsCollisionHand script has a few experimental modes:
image

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -2,6 +2,8 @@
 - Environments can now be set normally in scenes loaded through the staging system.
 - Fixed issue with not being able to push rigid bodies when colliding with them.
 - Fixed player movement on slopes.
+- Fixed lag in finger-poke.
+- Added initial collision hand support.
 
 # 4.1.0
 - Enhanced grappling to support collision and target layers
--- a/addons/godot-xr-tools/functions/movement_climb.gd
+++ b/addons/godot-xr-tools/functions/movement_climb.gd
@@ -43,14 +43,14 @@ const SNAP_DISTANCE : float = 1.0
 @export var velocity_averages : int = 5
 
 
-## Left climbable
-var _left_climbable : XRToolsClimbable
+# Left climbing handle
+var _left_handle : Node3D
 
-## Right climbable
-var _right_climbable : XRToolsClimbable
+# Right climbing handle
+var _right_handle : Node3D
 
-## Dominant pickup (moving the player)
-var _dominant : XRToolsFunctionPickup
+# Dominant handle (moving the player)
+var _dominant : Node3D
 
 
 # Velocity averager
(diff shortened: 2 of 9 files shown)


DigitalN8m4r3:
l(almost)gtm
it seems to fix the hand drifting away/ staying behind issue we currently have

currently what is not working good is when usign it with climbables, might just be the same issue i had at the beggining regarding collision layers but this needs some more investigating since being able to pull urself up/down using climbables is actualy one of the main selling points for the col hand

BastiaanOlij:
I'll find some time this week to test as well but it's looking good so far.

Digitals issue with the drift we'll need to think about whether we fix that first or whether we leave that for a future enhancement. As we're not actually grabbing anything and we're still attempting to follow the real hands position it's normal that the hand doesn't stay in one position.

The thing I wanted to add here is to have a mode that if a static body is grabbed, instead of anchoring the object to the hand (like we do with rigidbodies), we anchor the hand to the object. In this case it becomes a lot easier because we simply switch by following the parent (the `XRController3D` node) to "following" the thing we've grabbed (which in essence will mean that we disable collisions and just position the `XRToolsCollisionHandNode` on one of the grab points). 
If the position of the node gets to far away from the XRController3D node we'll "let go" automatically.

DigitalN8m4r3:
gettin better there,
climbing works pretty good now, only thing i dont like is when u climb the offset is too fast, introducing a smooth/ease would be verry welcome since it could otherwise intoduce motion sickness

collision on pickables, not good the way it is introduced in the pickables demo... the pickables collide just with the hands and upon drop perform a instant throw off due to collision issues with the other hand
(lets just keep this simple and plain and use actual pickable collision the way i intended it in the col demo)

bugs/issues:
snap turining while holding objects - object lags behind and adds a visual distortion due to frame inconsistency
moving while holding objects - object lags behind

BastiaanOlij:
> collision on pickables, not good the way it is introduced in the pickables demo... the pickables collide just with the hands and upon drop perform a instant throw off due to collision issues with the other hand
> (lets just keep this simple and plain and use actual pickable collision the way i intended it in the col demo)

@DigitalN8m4r3 we've simply not implemented this yet, Malcolm just added collision hands to all the demos just to test various things but that will be remove before we merge this.

We're looking at enhancing what you build to include other collision shapes as well, one of the things I want to combine for instance is the logic in physics hands (which will likely become obsolete), so this will likely all happen in a follow up PR

BastiaanOlij:
@Malcolmnixon this looks good to me for a first pass, works pretty well. As discussed we should remove the changes to the existing demos and rebase the PR so we can merge.

Follow ups then become (in no particular order):
- rebasing Digitals PR and getting most of his example scene in to demo this logic
- start working on implementing his collision improvements combined with our ideas to also combine in physics hands
- look into the two handed mode
- evaluate which of our current scenes should include collision hands

Something I also want to start thinking about is if we can do something with angular collisions instead of just re-orienting the hands. It be nice that if we bring a hand down on a table, it flattens itself on the surface instead of retaining the angle of the hand. Though I don't think we can start looking into that until after we have our finger collisions

BastiaanOlij (approved):
lgtm, if we merge I'll make some time tomorrow to do a bunch of git-fu to get more of digitals work in.

PR #491, opened 2023-07-30
Support XRToolsHand being any depth under an XRController3D

Malcolmnixon:
This pull request relaxes the need for an XRToolsHand to be a direct child of an XRController3D, and instead the XRToolsHand can be any depth in the tree under an XRController3D.

This allows for more complicated hand controller configurations such as collision hands.


BastiaanOlij (approved):
Thanks for splitting this out, lgtm, can be merged :)

PR #492, opened 2023-07-31
Mark project settings as basic so they are always shown

BastiaanOlij:
This is only possible in v4.1 onwards, but the way I did this should mean the change is just skipped in 4.0, but in order to see our project settings right away instead of selecting "advanced settings" is by marking them as basic.
We are doing that in this PR.


Malcolmnixon (approved):
Looks good and tested it on both Godot 4.0.3 (no set_as_basic) and 4.1.1 (has set_as_basic) and everything worked as expected.

Issue #488, opened 2023-07-26
Add a material template to viewport2d_in_3d   (enhacement)


surreal6:
Idea from [https://github.com/GodotVR/godot-xr-tools/pull/476#pullrequestreview-1542116766](https://github.com/GodotVR/godot-xr-tools/pull/476#pullrequestreview-1542116766)

```
A alternate approach you might want to consider would be 
to allow the user to optionally specify their own 'template' material. 

If not specified the viewport_2d_in_3d would make a new one as it does today, 
but if the user provides a template tweaked with custom settings, 
the viewport_2d_in_3d would duplicate it and then override the properties 
it needs to make the viewport work.

This would give the developer complete control over essentially all material 
properties - not just the few we've got and the depth/priority you've added.
```

Malcolmnixon:
This should be good to close now based on the pull request that went through.

BastiaanOlij:
Indeed, seeing its implemented, closing this issue :)

PR #493, opened 2023-08-02
Viewport 2D in 3D material cleanups

Malcolmnixon:
This pull request implements feature request #488 by:
- Allowing the user to specify a custom material for the screen
- Hide material overrides (alpha-scissor-threshold, unshaded, filtered) when using a custom material
- In the editor, refresh the screen material every second to show material updates performed in the property editor

Additionally other changes include
- Use a range slider for alpha_scissor_threshold to ensure [0..1] range
- Fixed unshaded option in Godot 4 via shading_mode
- Fixed editor tool failure when clearing scene
- Used dirty-flags approach to updating render objects to minimize unnecessary changes

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -4,6 +4,7 @@
 - Fixed player movement on slopes.
 - Fixed lag in finger-poke.
 - Added initial collision hand support.
+- Added support for custom materials for 2D in 3D viewport
 
 # 4.1.0
 - Enhanced grappling to support collision and target layers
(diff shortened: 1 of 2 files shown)


Malcolmnixon:
This method of letting the user provide their own material allows for fine control of the rendering - similar to pull request #476:

image

BastiaanOlij (approved):
lgtm, this should solve all current and future wishes around controlling the material.

We'll need to document it properly, especially that the albedo texture will always be set by us.

surreal6:
Hi @Malcolmnixon i'm sorry i have no time to checkout and test it, but i read the code and this solution seems good for me, also with this it's easier to understand the viewport workflow for everyone. 

You can also close my PR when you merge this 👍   Thanks!

PR #495, opened 2023-08-05
Digitals Boxing Ring collision hands demo

BastiaanOlij:
First in a couple of PRs that will add Digitals boxing ring demo (#478) into XR Tools while solving a number of rebase issues and adjust the demo to work with the logic we're polishing up.

I'm not finished with this yet as time is short but some early feedback would be appreciated.

Things that will be fixed/dealt with in follow up PRs:
- Digitals improvement that copies collision shapes when you grab an object is still missing from this (I'm also planning on adding collisions for fingers)
- Digitals improvement for reacting to the weight of the item being picked up is still missing from this
- When dropping an object it will often collide with the collision hands and go flying off, Malcolm is working on a fix for this
- The climbing wall will return, we're working on a different solution for the hand grabbing.
- Two handed pickup will return, this was still experimental even in Digitals PR and we want to separate it out so it doesn't become a blocking issue.


DigitalN8m4r3:
bump
from what i understand, this apparently will not be completed for 4.3 and we might want to move this to xr tools v2.

from my pov, i tried recreating the pickable collision logic with the New collision hand and it falls and overcomplicates, we would need to account for twohanded and for evry other extra handposition, this would mean to either move the collision copy evry time or to Delete and recreate it, this Producess unwanted physics reactions

Regarding weight, i was unable to add drag to the forcebody3d node, probably just my lack of skill here.

Amyways, this pr is soon havin its 1st birthday, so cake or no is the question

BastiaanOlij:
I'll rebase this again so that it is mergeable but leave out anything that relies on features that we haven't implemented in XR Tools 1.

In XR Tools 2 I've already implemented proper collisions on fingers and transfering collisions from picked up object, so we'll continue on the implementation there once public.

DigitalN8m4r3:
> I'll rebase this again so that it is mergeable but leave out anything that relies on features that we haven't implemented in XR Tools 1.
> 
> In XR Tools 2 I've already implemented proper collisions on fingers and transfering collisions from picked up object, so we'll continue on the implementation there once public.

Just to clarify, v1 wont get pickable collision? and were u by any Chance succesfull adding the weight feel?

BastiaanOlij:
I need to do a bit more testing but this is now rebased on the current collision hands logic so most should be working.

BastiaanOlij:
Definitely have a few more things to fix. Can't walk up the stairs, obviously weights don't work (yet) and hitting the punching bag needs to be improved a little

DigitalN8m4r3:
> Definitely have a few more things to fix. Can't walk up the stairs, obviously weights don't work (yet) and hitting the punching bag needs to be improved a little

Will test it soon, weights? that would complement v1..
Anyways good work!

PR #497, opened 2023-08-05
Handle scaling of Viewport2Din3D

Malcolmnixon:
There are times when a Viewport2Din3D has to be scaled - such as when the screen is worn by the player, and the player changes size. This pull request makes Viewport2Din3D scaling functional by:
- Using affine_inverse for global_to_viewport to handle scaling
- Move the collider from "around" the screen to "behind" the screen

The normal `Transform3D.inverse()` method only works when the transform has no scaling. When the transform has scaling the `Transform3D.affine_inverse()` must be used instead.

The change to the collider was because the collider surrounded the screen (1cm in front and back). While this isn't too bad when using an unscaled Viewport2Din3D, the problem gets worse as the screen is scaled up.
image


BastiaanOlij (approved):
lgtm

PR #502, opened 2023-08-08
Enhance pointer function

Malcolmnixon:
This pull request enhances the pointer function to support more customizability. The following are the changes:
- Added new signals to function pointer.
- Grouped function pointer properties.
- Added laser material and hit-material properties.
- Added target material and radius properties.
- Modified the virtual keyboard so its viewport properties are accessible and defaulted to opaque/unshaded.
- Added custom pointer settings to the demo.

The function pointer properties are grouped as follows:
image

Expanding them shows the full set of properties:
image

The following is a screen-shot of the pointer with customizations in the demo scene:
image

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -5,6 +5,8 @@
 - Fixed lag in finger-poke.
 - Added initial collision hand support.
 - Added support for custom materials for 2D in 3D viewport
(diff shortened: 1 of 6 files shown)


BastiaanOlij (approved):
Nice!

PR #503, opened 2023-08-10
Improve control over pickable object highlighting

Malcolmnixon:
This pull request switches the pickable object highlighting methods from:
```gdscript
func increase_is_closest()
func decrease_is_closest()
```
to:
```gdscript
func highlight_request(from : Node, on : bool)
```

The old method used reference counting which was prone to dropping counts, and was intended only for proximity-triggering.

The new method saves tracks which nodes are asking for the highlight, so a node can safely set or clear its request any number of times without interfering with other requests. Additionally the highlighting could be for any reason not just proximity.

Finally this method now allows snap-zones to forward highlighting requests to objects in the snap zones. This makes it easier to see which object is ready to be pulled out of a snap-zone.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -8,6 +8,7 @@
 - Updated pointer to support visibility properties and events
 - Modified virtual keyboard to expose viewport controls and default to unshaded
 - Cleaned up teleport and added more properties for customization
+- Modified pickup highlighting to support pickables in snap-zones
 
 # 4.1.0
 - Enhanced grappling to support collision and target layers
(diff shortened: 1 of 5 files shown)


jkankiewicz:
@Malcolm Nixon, thank you!
After installing PR #503 and fixing  bug #505 (PR #506), an `XRToolsPickable` node attached to a `XRToolsSnapZone` node highlights when an `XRController3D` node is close to it just like it does when it's unattached.

BastiaanOlij (approved):
Should have an entry in changes doc, other than that good to merge

PR #504, opened 2023-08-10
Teleport improvements

Malcolmnixon:
This pull request cleans up the teleport function by:
- Changing it to extend from a Node3D rather than the CharacterBody3D workaround
- Grouped and enhanced properties

The player can now control the teleport visual textures and the material of the player capsule.

image
image


Malcolmnixon:
Added an option to set a scene as the players body.
image

BastiaanOlij (approved):
Very cool!

Issue #505, opened 2023-08-10
Invalid get index 'playing' (on base: 'null instance') at addons/godot-xr-tools/objects/snap_zone.gd:248


jkankiewicz:
My Godot-4.1.1 project doesn't play a sound when snapping an `XRToolsPickable` into an `XRToolsSnapZone` and it crashes within `pick_up_object()` because an `AudioStreamPlayer3D` node is expected to exist.

Malcolmnixon:
Could you test with the latest master branch, and if it's fixed we can close the issue.

BastiaanOlij:
Closing this as fixed.

PR #506, opened 2023-08-10
Added check for a valid `AudioStreamPlayer3D` node.

jkankiewicz:
This PR fixes issue #505:
* Added check for a valid `AudioStreamPlayer3D` node.

### DIFF
--- a/addons/godot-xr-tools/objects/snap_zone.gd
+++ b/addons/godot-xr-tools/objects/snap_zone.gd
@@ -249,10 +249,11 @@ func pick_up_object(target: Node3D) -> void:
 	# Pick up our target. Note, target may do instant drop_and_free
 	picked_up_object = target
 	var player = get_node("AudioStreamPlayer3D")
-	if player.playing:
-		player.stop()
-	player.stream = stash_sound
-	player.play()
+	if is_instance_valid(player):
+		if player.playing:
+			player.stop()
+		player.stream = stash_sound
+		player.play()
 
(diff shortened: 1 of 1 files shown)


Malcolmnixon:
The original code appears to be written assuming users will:
- Create their own snap-zone scene by extending from snap_zone.tscn, or enabling "Editable Children"
- Edit the AudioStreamPlayer3D to configure the (numerous) sound/bus/volume settings

The original code isn't so friendly to instantiation via "Add New Node" as it requires the user create the CollisionShape3D and the AudioStreamPlayer3D. This PR relaxes this slightly by allowing the user to omit creating the AudioStreamPlayer3D if they don't want sounds.

After we get this PR in, I think we should continue to improve the "Add New Node" behavior by:
- Allowing the user to provide their own CollisionShape3D
- If the user doesn't provide their own CollisionShape3D then programmatically creating a sphere CollisionShape3D and exposing the Grab Distance property
- Allowing the user to provide their own AudioStreamPlayer3D
- Hiding the "Stash Sound" if the user doesn't provide their own AudioStreamPlayer3D

Note: It's simple for us to create a sphere CollisionShape3D programmatically; but we don't want to create an AudioStreamPlayer3D as there are so many audio properties - it's far easier to let the user manipulate via having them put in an AudioStreamPlayer3D node.

Malcolmnixon (approved):
This looks good

BastiaanOlij (approved):
lgtm

PR #509, opened 2023-08-21
Added passing of data between staged scenes

Malcolmnixon:
This pull request allows optional data to be passed between scenes. The default handler of this data will use it to try and override the spawn-point of the player. The specifics of this pull request are:
- Added passing of data between staged scenes
- Added passing of user_data between scenes.
- Added default handling of string/vector3/transform3d user_data as spawn-point.
- Added extended documentation about staging signals and methods.
- Modified demo teleporter to support optional spawn-point data.

### DIFF
--- a/addons/godot-xr-tools/staging/scene_base.gd
+++ b/addons/godot-xr-tools/staging/scene_base.gd
@@ -2,32 +2,41 @@
 class_name XRToolsSceneBase
 extends Node3D
 
-## Introduction
-#
-# This is our base scene for all our levels.
-# It ensures that we have all bits in place to load
-# our scene into our staging scene.
 
-
-## Request staging exit to main menu
+## XR Tools Scene Base Class
 ##
+## This is our base scene for all our levels.  It ensures that we have all bits
+## in place to load our scene into our staging scene.
+##
+## Developers can customize scene transitions by extending from this class and
+## overriding the [method scene_loaded] behavior.
+
+
 ## This signal is used to request the staging transition to the main-menu
 ## scene. Developers should use [method exit_to_main_menu] rather than
 ## emitting this signal directly.
 signal request_exit_to_main_menu
 
-## Request staging load a new scene
-##
 ## This signal is used to request the staging transition to the specified
 ## scene. Developers should use [method load_scene] rather than emitting
 ## this signal directly.
-signal request_load_scene(p_scene_path)
-
-## Request staging reload the current scene
 ##
+## The [param user_data] parameter is passed through staging to the new scenes.
+signal request_load_scene(p_scene_path, user_data)
+
 ## This signal is used to request the staging reload this scene. Developers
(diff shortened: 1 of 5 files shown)


addons/godot-xr-tools/staging/scene_base.gd
```diff
@@ -60,28 +64,74 @@ func center_player_on(p_transform : Transform3D):
...
-	center_player_on($XROrigin3D.global_transform)
+	# Inspect standard user data types to override the players starting location
+	# in a scene. Overriding this behavior can be achieved by overriding this
+	# function and calling super() with:
+	# - null to use the standard XROrigin3D location
+	# - string name of a Node3D to spawn at
+	# - Vector3 to spawn at
+	# - Transform3D to spawn at
+	var spawn_position : Transform3D = $XROrigin3D.global_transform
+	match typeof(user_data):
```
BastiaanOlij, line 94: I would add a `TYPE_OBJECT` in there as well, and check for say a method `get_spawn_position`, this way you could parse a "game data object" around that contains persistent game data from scene to scene. 

(or maybe we create a base class for this in XR Tools?)

addons/godot-xr-tools/staging/scene_base.gd
```diff
@@ -60,28 +64,74 @@ func center_player_on(p_transform : Transform3D):
...
-func scene_visible():
+	# Center the player on the spawn location
+	center_player_on(spawn_position)
+
+
+## This method is called when the scene becomes fully visible to the user.
+##
+## The [param user_data] parameter is an optional parameter passed in when the
+## scene is loaded - usually from the previous scene.
+func scene_visible(_user_data = null):
```
BastiaanOlij, line 114: I'm assuming you added the `_` prefix to shut the validation errors of the code parser up?

Seeing this is an actual parameter I'd like to remove this prefix. You can add a:
`@warning_ignore("unused_parameter")`
before the function definition.

BastiaanOlij (approved):
Looking good, just some nitpicking issues.

My only worry is breakage on `XRToolsSceneBase.scene_loaded`. If someone has implemented their own code in an extended scene, GDScript will not call that because it hasn't got the `user_data` parameter now defined in it.

We should probably add a `minor-breakage` tag in the `VERSIONS.md` file.

We probably should think of how we want to maintain small footnotes about this in our releases (and since I copy those from `VERSIONS.md` maybe there as well) to instruct users to make minor changes like this.

addons/godot-xr-tools/staging/scene_base.gd
Malcolmnixon: This attribute isn't currently supported by gdlint, but instead I added:
```gdlint
# warning-ignore:unused_parameter
# gdlint:disable=unused-argument
```

This handles both Godot and gdlint warnings.

Malcolmnixon:
I added a bolded minor-breakage at the start of the VERSIONS.md line to highlight the change.

PR #510, opened 2023-08-22
Added process_physics_priority settings for Godot 4.1+

Malcolmnixon:
This pull request adds the `process_physics_priority` settings to objects needing custom physics order. The settings are required for Godot 4.1+, but are ignored for Godot 4.0 which uses the `process_priority` setting for both render and physics.


BastiaanOlij (approved):
Cool, so we'll need to be careful that we don't accidentally undo these as opening up the project in Godot 4.0.3 will eventually clear them out when it decides to (re)save these scenes.

Issue #499, opened 2023-08-07
Feature request: Provide more signals


tcrass:
I stumbled over this when I was trying to implement the following behaviour: I want a controller's teleport function to be disabled while the same controler's pointer function is hitting a target. However, I currenly don't see any possibility to determine from "outside" whether the pointer is pointing at a pointable or not -- there seems to be neither a property reflecting the pointer's state, nor (imho the preferred mechanism) a signal notifying about the pointer's state.
Furthermore, I'd like to play different sounds on various function events, like pointer (de-)activation, pointer hitting something, teleporter (de-)activation, teleporter target state change, and actual teleportation. Again, it would be great if signals were available for notifying about all these things.
I suppose it would be generally appreciated if all functions provided signals regarding their specificv state changes.

Malcolmnixon:
The pointer function exposes the 'last_target' and 'last_collided_at' fields; however it's inefficient to poll them so pull request #502 adds the signals I think you're looking for:
- `pointer_entered(target)`
- `pointer_moved(target, from, to)`
- `pointer_pressed(target, at)`
- `pointer_released(target, at)`
- `pointer_exited(target)`

tcrass:
OK, I'm on it... My current version of the iterface looks like

```
teleporter_activated()
can_teleport_changed(can_teleport)
teleported_to(location)
teleporter_deactivated()
teleporter_target_entered(node, at)
teleporter_target_moved(node, from, to)
teleporter_target_exited(node, at)
```

I've so far implemented the first four signals, but for the last three (the target-related ones) I'll first have to gain a deeper understanding of the current teleport function code as well as Godot 4's physics system. Might happen tomorrow, might as well take another month... ;)

Malcolmnixon:
Pull request #508 moves the teleport player-movement logic into XRToolsPlayerBody, and the player emits a player_teleported signal after the player moves. This is because the new XRToolsTeleportArea can also teleport a player when the user steps into it.

BastiaanOlij:
@tcrass were you planning on submitting a PR for those signals? Looks like a worth while improvement of the teleport logic.

tcrass:
> were you planning on submitting a PR for those signals? Looks like a worth while improvement of the teleport logic.

I'm on it: https://github.com/tcrass/godot-xr-tools/tree/499-teleport-signals

[...]

PR #513, opened 2023-09-02
More signals for teleporter function

tcrass:
Implementation of the new teleporter signals suggested in #499 

One thing, though, I'm not sure about is that after merging my feature branch back into master there was one line (line 21, `player_material = ExtResource("4")`) missing from `function_teleport.tscn`. I don't think I touched the scene file, but I can't explain to myself how, when and where that line went, either.

### DIFF
--- a/addons/godot-xr-tools/functions/function_teleport.gd
+++ b/addons/godot-xr-tools/functions/function_teleport.gd
@@ -12,6 +12,28 @@ extends Node3D
 ## a teleport function on that controller.
 
 
+## Signal emitted when the teleporter is activated
+signal teleporter_activated()
+
+## Signal emitted when the teleporter's ability to perform a teleportation changes
+signal can_teleport_changed(can_teleport)
+
+# signal emitted when the teleporter beam hits a suitable target
+signal teleporter_target_entered(node, at)
+
+# signal emitted when the teleporter beam moves along a suitable target
+signal teleporter_target_moved(node, from, to)
+
+# signal emitted when the teleporter beam leaves a suitable target
+signal teleporter_target_exited(node, at)
+
+## Signal emitted when the player gets teleportet to the teleporter's target location
+signal teleported_to(location)
+
+## Signal emitted when the teleporter is deactivated
+signal teleporter_deactivated()
+
+
 # Default teleport collision mask of all
 const DEFAULT_MASK := 0b1111_1111_1111_1111_1111_1111_1111_1111
 
@@ -84,8 +106,10 @@ var player_material : StandardMaterial3D = _DefaultMaterial :  set = set_player_
 
 
 var is_on_floor : bool = true
-var is_teleporting : bool = false
-var can_teleport : bool = true
+var is_teleporting : bool = false: set = set_is_teleporting
+var can_teleport : bool = true: set = set_can_teleport
+var target_node : Node3D = null
+var target_hit_pos : Vector3
 var teleport_rotation : float = 0.0;
 var floor_normal : Vector3 = Vector3.UP
 var last_target_transform : Transform3D = Transform3D()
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
I think the removal of the player material is fine, this probably is a result of the IDE overriding the player material when it was introduced, and now that it's reloaded everything finding out that the overridden material and the embedded material are the same and thus not needed.

BastiaanOlij (approved):
Looks good to me, the only thing I'm wondering about is whether it will be inconsistent in sending out signals at startup when it assigns default value. But I doubt that will be an issue in real life.

tcrass:
Hi, I just saw that signalling has been re-designed for the pointer function. The pointer now only provides one single signal, featuring an enum telling what kind of pointer event happened. This is, in fact, very similar to something I initially also came up with for the transporter function, but which I decided against for sake of consistency with what back then was the pointer's way of doing things. What do you think -- wouldn't it now be appropriate to re-work transporter signalling to also provide only a single signal sending "meta-events"?

Cheers -- tcrass

Malcolmnixon:
The pointer_event is useful because pointers need to encapsulate large amounts of information (what is pointing, who it's pointing at, what the event was, and what the 3D world location is of the contact). As such it pretty much has to be interpreted by code to translate all this information into something meaningful.

Changing the teleport signals to a single combined object prevents using the signal-connection inspector from wiring up discrete action. In cases where you want to just play a sound in reaction to teleport-activated or teleport-exited, or teleported; you would now require a script to break apart the single event and inspect the reason.

tcrass:
> The pointer_event is useful because pointers need to encapsulate large amounts of information (what is pointing, who it's pointing at, what the event was, and what the 3D world location is of the contact). As such it pretty much has to be interpreted by code to translate all this information into something meaningful.
> 
> Changing the teleport signals to a single combined object prevents using the signal-connection inspector from wiring up discrete action. In cases where you want to just play a sound in reaction to teleport-activated or teleport-exited, or teleported; you would now require a script to break apart the single event and inspect the reason.

Four out of seven transporter events also provide information about "what" is hit and "where" it is hit -- and on the other hand, in my use-case, I'm only interested in whether the pointer enters or exits something, but I don't care what and where. So from my point of view, pointer and teleport eventing isn't that different at all...

Anyway, I've already re-worked the teleporter script to provide a single teleport event, similar to the pointer case. I'll submit another pull request for that, and you go ahead and just pick whatever you consider more suitable.

Cheers -- tcrass

tcrass:
Seems I still haven't understood how pull requests work on GitHub -- I expected a pull request to always refer to a specific commit, but I just realized that the latest changes I made in my fork's master branch got automatically added to this pull request. This is not what I intended, sorry... No I'm also unsure whether I should create pull requests from my master branch (after updating it to the latest state of your repository), or if it would be better to create individual feature branches for different pull requests.

Anyway, regarding the transporter event implementation: Just let me know if you prefer the "multiple events" or the "all-in-one event" version. In case of the former, I'll just revert the commit implementing the latter, ok?

Cheers -- tcrass

Malcolmnixon:
> nderstood how pull requests work on GitHub -- I expected a pull request to always refer to a specific commit, but I just realized that the latest changes I made in my fork's master branch got automatically added to this pull request. This is not what I intended, sorry... No I'm also unsure whether I should create pull requests from my master branch (after updating it to the latest state of your repository), or if it would be better to create individual feature branches for different pull requests.
> 
> Anyway, regarding the transporter event implementation: Just let me know if you prefer the "multiple events" or the "all-in-one event" version. In case of the former, I'll just revert the commit implementing the l

Yeah, the general approach is to make a branch in your forked repository for any one feature, and then construct the pull request from that branch. That way you can iterate and clean up the PR in isolation.

I think it's best to keep it as multiple events. For similar reasons I'm also going to add a few helper signals for #537.

BastiaanOlij:
This needs updating for the current version of XR Tools if we are going to merge this.

We are planning a brand new teleporter implementation for XR Tools 2 so may take this along.

PR #514, opened 2023-09-03
Fixed pickup function signals

BastiaanOlij:
Connecting to signals was assuming our parent is the controller, this is no longer true as we can have our collision hands in the middle.

Now connecting to our controller.


Malcolmnixon (approved):
Looks good

PR #515, opened 2023-09-03
Fix issues with static body not being detected by area3D

BastiaanOlij:
Turns out that `StaticBody3D` is no longer detected by `Area3D` in Godot 4.
This is causing issues when using collision hands and interactable.

One workaround seems to be to use the new `AnimatableBody3D`.

Still looking into other effected areas.


Malcolmnixon (approved):
Looks good and plays well

BastiaanOlij:
So far this seems to indeed work as intended.

PR #516, opened 2023-09-05
Added popup interface support

Malcolmnixon:
This pull requests adds support for popup menus in game. It includes:
- Adding XRToolsPopup script/scene for managing popup interfaces.
- Adding XRToolsFunctionPointer static find helper methods.
- Adding class_name to XRToolsViewport2Din3D.
- Modifying XRToolsViewport2Din3D to only enable when visible.
- Adding layer 24 - Popup UI
- Adding popup menu to pointer-demo scene

image

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -2,6 +2,7 @@
 - Upgraded project to Godot 4.1 as the new minimum version.
 - Added reporting of stage load errors.
 - Blend player height changes and prevent the player from standing up under a low ceiling.
+- Added popup support for game menus.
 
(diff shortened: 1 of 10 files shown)


DigitalN8m4r3:
lgtm, like the idea of this but yeah i heard some concerns about its implementation, someone suggested doing a raycast i think

anyways this works good

BastiaanOlij:
Discussed with Malcolm, we are going to rethink this approach, especially with the new composition layers being a better fit for menus. Closing this and will raise something new in the future.

Issue #384, opened 2023-02-23
Player crouch should work better with colliders


Malcolmnixon:
Currently when the player starts or stops crouching, the player body instantly snaps to the new height. Shrinking isn't usually a problem, but growing the player body instantly to be over 2x the height can actually push more than half the players body beneath the floor causing the player to fall through the floor.

Additionally the player body does not check whether there is room to grow, so the player is able to stand up inside a cave. Instead the body should check (probably by move_and_collide test) to see how far the player can grow.

DigitalN8m4r3:
indeed, now it makes sense why i ended up havin always issues in my hidemechanics, as they are built on the idea of scalin the player to crouch and back to standing. Good catch

Issue #480, opened 2023-07-18
godot-xr-tools/addons/godot-xr-tools/functions/movement_crouch.gd: changes PlayerBody height from bottom down causing it to stick on release in thin colliders


GeminiSquishGames:
When using thin mesh or terrain colliders, (eg, height map shape, Zylann's hterrains or Zylann's voxel terrains module), and probably general concave collision shapes as well, the PlayerBody's collider is snapped from it's bottom in one frame down into the thin collider when .clear() is applied to the dictionary _player_height_overrides. So when the crouch is released the PlayerBody is stuck rather than standing. This should probably be done in a way that changes height from the top so it doesn't snap through these colliders. Preferably smooth. This is observed in #384 but not very well explained.

update: After looking at things, I'm not certain, but it could probably be fixed better with more work on the Godot Physics engine rather than the XRTools. The colider being stuck doesn't seem to be something that should happen at a lower level maybe? I will try and look into that as well as trying another physics engine if I can find one with a module. Still will have to work around it some how since none of my outdoor levels will work with it.

update: My work around, though probably a terrible idea:
@ line 74:
`
#Update crouching state

    if crouching != _crouching:
        _crouching = crouching        
        if crouching:
            player_body.override_player_height(self, crouch_height)
        else:                  
            player_body.override_player_height(self)
            player_body.position.y += crouch_height # added to pop out of certain colliders`

It "seems" to work with the height map collider node, and Zylann's hterrains and voxel terrains collisions, but needs peer review and further investigation I feel.

PR #519, opened 2023-09-08
Add smooth player height control

Malcolmnixon:
This pull request cleans up crouching and fixes issues #384 and #480 by:
- Smoothly blending between player-controlled-height and software-override-height.
- Block player-height from growing when blocked by a low ceiling.

The following video demonstrates its behavior in the test scenes - https://youtu.be/D43q14dIUKc

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,7 @@
 # 4.3.0
 - Upgraded project to Godot 4.1 as the new minimum version.
 - Added reporting of stage load errors.
+- Blend player height changes and prevent the player from standing up under a low ceiling.
 
 # 4.2.1
 - Fixed snap-zones showing highlight when disabled.
--- a/addons/godot-xr-tools/player/player_body.gd
+++ b/addons/godot-xr-tools/player/player_body.gd
@@ -65,6 +65,9 @@ const NEAR_GROUND_DISTANCE := 1.0
 ## Maximum player height
 @export var player_height_max : float = 2.5
 
+## Slew-rate for player height overriding (button-crouch)
+@export var player_height_rate : float = 4.0
+
 ## Eyes forward offset from center of body in player_radius units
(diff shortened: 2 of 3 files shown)


DigitalN8m4r3 (approved):
**_Tested this in and out as it delivers a much needed feature that the XR Tools are lacking_**

*a way to prevent the players head from moving thru ceilings and blockade spots

Now this PR adds the above and some more as it accounts for the physical height change when the player suddenly stands up, the effect ingame is that the player is still under the blockade spot. 

I might be biased as i was working on and off on a way to do just that but was to much locked in the idea of inducing a motion sickness way to do this, however since i had the chance to test this implementation and compare it to similar ideas in unity, i can say that this is it. It works and accounts for sitting/standing up and vice versa without breaking the immersion.

One thing i would like to request would be to add this to the possible 2.1 HotFix as it paths the way for other things such as the Seating PR #462 to work in a more homogeneous way

pietru2004:
I don't see much problems althrou the hand gui could have box for making finger straight, I keep bumping into my hand with it nor straight, well I guess it is not from this merge req...
Also I couldn't see problems with height ovverride change and the low celing worked and did not make my cam go throu it

pietru2004 (approved):
👍

BastiaanOlij (approved):
lgtm, but needs testing!

Malcolmnixon:
Added a YouTube video showing the new behavior.

GreenboxGames:
I can confirm this fixes my issue #480 with both plane collisions and terrain collisions. However, if I jump while uncrouching, it seems to have the side effect of jump boosting when you jump while uncrouching. This could be a feature, but not in all cases of controller.

Malcolmnixon:
I tracked down the jump-boosting to a double-jump and prevented it with a better "standing on the ground" test.

DigitalN8m4r3:
lgtm, no double jump noticed :)

Issue #271, opened 2022-11-16
Feature request - world-grab movement


Malcolmnixon:
This movement provider would allow the user to "grab" the ether/space-time and move/rotate/scale themselves.

PR #520, opened 2023-09-09
Add world-grab movement support

Malcolmnixon:
This pull request implements feature #271 by adding a world-grab movement system.

The XRToolsMovementWorldGrab works with standard XRToolsFunctionPickup nodes on the players hands to support grabbing XRToolsWorldGrabArea regions of the scene.


DigitalN8m4r3:
forgot to actualy do a rewiew here but i had time to test this and besides adding the above it does bring in the question if this could be used to do world building in vr

anyways back to the actual pr, it works so far good and does not seem to add any issues
lgtm

BastiaanOlij (approved):
Code wise this looks fine, would be nice to get some more feedback from people who tried it out whether there is anything we should improve.

Issue #522, opened 2023-09-12
Snap-zone highlighting should be suppressed when disabled


Malcolmnixon:
Currently snap-zones highlight when empty regardless of their `enabled` property. Instead they should only highlight when both empty and `enabled`.

PR #523, opened 2023-09-12
Fix snap-zones highlighting when disabled

Malcolmnixon:
This pull request fixes issue #522 by using the `enabled` property when sending the highlight_updated and close_highlight_updated signals. A disabled snap-zone always requests highlighting be disabled.


BastiaanOlij (approved):
lgtm

PR #528, opened 2023-09-16
Update docs.yml

The-Cyber-Captain:
Remove reference to missing [deleted?] page - "player" -  so nearby Next / Prev functions don't error 404


Malcolmnixon (approved):
I tested this out locally and it works fine. Thanks!

Issue #529, opened 2023-09-17
Collision hands should snap to the real hands if too far away


Malcolmnixon:
If the collision hands get separated too far from the real hands, then they should snap back to the real hands.

PR #530, opened 2023-09-17
Fix collision hands when they get too far from the real hands

Malcolmnixon:
This pull request fixes issue #529 by detecting when the collision hands get too far from the target, and teleporting them.


BastiaanOlij (approved):
Ah must have missed that one, I believe that was in Digitals original implementation as well.

PR #532, opened 2023-09-21
enable OptionButton to work from vr-simulator

goatchurchprime:
When trying to click on an OptionButton using the [xr-simulator](https://github.com/Cafezinhu/godot-vr-simulator) add-on they never popup.  This seems to be due to it receiving double clicks.  

I don't know exactly what this push_event function is doing as a lot of events go through it that don't get into the UI (when not holding down the E key in the simulator), but I think it's responsible for duplicate clicks going through and immediately closing the popup.  

There's something problematic happening in the implementation, as reported at: https://github.com/godotengine/godot/issues/81579#issuecomment-1729468644

(The other bug where the popup remains activated and pops up again if you had selected the option in the mode of a click release is still there after this change.)

### DIFF
--- a/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
+++ b/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
@@ -199,7 +199,8 @@ func _on_pointer_event(event : XRToolsPointerEvent) -> void:
 
 # Handler for input eventsd
 func _input(event):
-	$Viewport.push_input(event)
+	if not (event is InputEventMouseButton):
+		$Viewport.push_input(event)
 
 
 # Process event


BastiaanOlij:
Godot 4's new multiple window support broke a number of things around viewports and events properly being handled because a lot of the UI elements assume subviewports are presented through subviewportcontainers and are thus purely 2D. 
The GUI in 3D logic has become very dodgy, see also: https://github.com/godotengine/godot/issues/77189

That said, in XR obviously you don't have a lot of input to worry about, we're using controllers that emulate mouse actions directly on screen. The simulator breaks that so there are limitations to what we can fix without breaking our primary goal, logic working in XR. As an alternative, try out the new XR simulator Meta released so you properly emulate OpenXR without needing a headset.

That all said, I'm happy to accept this fix, we're not expecting mouse input, and the mouse emulation is handled inside of the component. If filtering out mouse events takes away the pain, everyone wins. I would recommend adding a good comment in the code to explain why mouse events are being filtered out.

BastiaanOlij (approved):
lgtm

Issue #535, opened 2023-09-22
Snap Zone grab distances incorrectly updating


Malcolmnixon:
In snap_zone.tscn the sphere collision-shape does not specify "Local To Scene", so the one shape resource is shared between all snap-zone instances. Therefore the last instance to set its grab_distance (the radius of the sphere) overrides all other grab distances.

PR #536, opened 2023-09-22
Fix shared grab distance

Malcolmnixon:
This pull request fixes issue #535 by changing the collider sphere-shape to be local-to-scene so it can be modified in each instance of the snap-zone.


BastiaanOlij (approved):
lgtm

Issue #542, opened 2023-09-26
Freeing a held object keeps the players hand in the old pose


Malcolmnixon:
The XRToolsPickable object can be configured with hand poses. If the XRToolsPickable is freed before dropping then the pickable doesn't get a chance to unregister its pose override on the hand.

This can be fixed by adding the following the the pickable.gd script:
```gdscript
# Called when the node exits the tree
func _exit_tree():
	# Revoke any hand pose
	if is_instance_valid(by_hand):
		by_hand.remove_pose_override(self)
```

PR #543, opened 2023-09-26
Fix hand pose on held object destruction

Malcolmnixon:
This pull request fixes issue #542 by having the pickable object unregister any hand-pose override when it leaves the scene-tree.


BastiaanOlij (approved):
lgtm!

Issue #547, opened 2023-10-08
Objects deleted from snap-zones can fail on highlight errors


Malcolmnixon:
The XRToolsSnapZone::request_highlight can fail if the object is freed with an "Attempt to call function 'request_highlight' in base 'previously freed' on a null instance."

PR #548, opened 2023-10-08
Add is_instance_valid check for highlighting item in snap-zone

Malcolmnixon:
This PR fixes issue #547 by adding an is_instance_valid check before attempting to highlight an item in a snap-zone in case it has been freed.


DigitalN8m4r3:
lgtm

BastiaanOlij (approved):
There are probably other places where this would be a worth while fix...

Malcolmnixon:
> There are probably other places where this would be a worth while fix...

I did a cursory check in this module and similar functions, and couldn't find any other occurrences... we'll have to see what pops up.

Issue #541, opened 2023-09-25
Grapple Interaction Collision Detection Gridmap Error


AinsRuby:
Hi, I'm attempting to use a Gridmap with collision on the tiles in my scene, due to the ease of use, and the grappling keeps crashing the project due to this line of code.
`var target : CollisionObject3D = _grapple_raycast.get_collider()`
Wondering if it would be possible to somehow get it working with a gridmap, or if I have to just use normal objects in my scene.
Any help would be appreciated!

Malcolmnixon:
What version of Godot are you using? I encountered something similar recently and it was caused by the `_grapple_raycast.is_colliding()` returning true but `_grapple_raycast.get_collider()` returned null.

Godot 4.1.x supports priorities on physics-processing, and it might be something to do with that. We mess with priorities to prevent latency issues with items in the players hands, but it may cause the raycast to not update. I'll look into this issue shortly.

Malcolmnixon:
My problem was caused by trying to get the RayCast3D to return a Node that had just been destroyed. I suspect your problem is that a GridMap does not extend from CollisionObject3D. Could you try the attached script and see if it works for your case?

[movement_grapple.gd.zip](https://github.com/GodotVR/godot-xr-tools/files/12722497/movement_grapple.gd.zip)

PR #551, opened 2023-10-17
Fix grappling on gridmaps

Malcolmnixon:
This PR fixes issue #541 by only updating the RayCast3D when needed in the physics frame (to prevent getting stale data), and by relaxing the type of collider to be of anything the RayCast3D can collide with - including GridMaps.

The following screen-shots were taken using the KayKit City Builder Bits assets in a GridMap.
image
image

### DIFF
--- a/addons/godot-xr-tools/functions/movement_grapple.gd
+++ b/addons/godot-xr-tools/functions/movement_grapple.gd
@@ -117,13 +117,21 @@ func _ready():
 	_line.hide()
 
 
-# Update grapple display objects
-func _process(_delta: float):
+# Update the grappling line and target
+func _physics_process(_delta : float):
 	# Skip if running in the editor
 	if Engine.is_editor_hint():
 		return
 
-	# Update grapple line
+	# If pointing grappler at target then show the target
+	if enabled and not is_active and _is_raycast_valid():
+		_grapple_target.global_transform.origin = _grapple_raycast.get_collision_point()
+		_grapple_target.global_transform = _grapple_target.global_transform.orthonormalized()
+		_grapple_target.visible = true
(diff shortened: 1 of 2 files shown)


DigitalN8m4r3:
Lgtm, 
on a sidenote, this could work well for a spiderman game :)

BastiaanOlij:
I would suggest trying to run everything in physics process and seeing if it runs properly, I'm not a fan of force physics because it takes a costly path especially when physics runs in a separate thread.

Malcolmnixon:
> I would suggest trying to run everything in physics process and seeing if it runs properly, I'm not a fan of force physics because it takes a costly path especially when physics runs in a separate thread.

Done, and tested as working in the same city map.

addons/godot-xr-tools/functions/movement_grapple.gd
```diff
@@ -134,14 +134,27 @@ func _process(_delta: float):
 		_line.visible = false
 
 	# Update grapple target
-	if enabled and !is_active and _is_raycast_valid():
-		_grapple_target.global_transform.origin  = _grapple_raycast.get_collision_point()
+	if enabled and !is_active and is_instance_valid(hook_object):
```
BastiaanOlij, line 137: Wouldn't it make more sense to move this into _physics_process in its entirety?

Malcolmnixon: Yeah, I was trying to keep rendering and physics updates separate; but if/when we do support running physics in a separate thread this will be the least of our concerns. I've merged them together into _physics_process.

BastiaanOlij (approved):
Looks good to merge

PR #558, opened 2023-10-31
Add double click on function_pointer

surreal6:
The idea is to add double click actions (i.e.: to allow navigation in a FileDialog, that requieres double click to enter in a folder)

I added two different ways to solve:

- one to detect double click in the trigger with an exposed debounce time

- another by declaring a button to the complete double click action (safer for reduced mobility, as double clicking in the air is a bit dificult for some people)    This second option is ignored if the value is left empty.

Let me know what you think :)

### DIFF
--- a/addons/godot-xr-tools/events/pointer_event.gd
+++ b/addons/godot-xr-tools/events/pointer_event.gd
@@ -10,6 +10,9 @@ enum Type {
 
 	## Pointer pressed target
 	PRESSED,
+	
+	## Pointer double click target
+	DOUBLECLICK,
 
 	## Pointer released target
 	RELEASED,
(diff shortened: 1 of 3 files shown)


Malcolmnixon:
The implementation looks reasonable. I wonder whether we should add something similar to poke.gd so people can trigger a double-tap with their finger. The duration would probably need to be a little longer due to no physical surface to use; although with the new haptic-system coming in it'll be easier to "feel" the touch and repeat a second one.

BastiaanOlij:
@Malcolmnixon to revisit this, looks good in general, we do feel the timer solution needs to change because there are situations this will fall over.

PR #563, opened 2023-11-07
Added implementation of two-handed grabbing

Malcolmnixon:
This PR implements two-handed grabbing. This is a work in progress.


DigitalN8m4r3:
nice, good start
obviously u know whats not working properly as in the actual snap of the secondary grab/hand but other then that this is verry promissing!

Malcolmnixon:
I did some fairly large refactoring including adding the position/angle/aim controls on hand-poses. The fact Godot-XR-Tools uses the "aim" pose rather than the "grip" pose is annoying. The following is a video showing what would be possible if we modified all the assets to work with the "aim" pose - note that the gun correctly rotates in the center of the primary hand.

https://github.com/GodotVR/godot-xr-tools/assets/1863707/398ca9ad-a1b5-4ab4-8045-97105b70890a

DigitalN8m4r3:
alright here it goes:

1. *Upon Drop the positioning should be smoothed out/ lerped as it is a bit distracting and feels more like a springjoint.

- purely a visual feedback thing


2a. *Drop while holding with Secondary should not make/force Secondary to be primary in position.
- On that note, when setting the Secondary to be a General one, upon drop of Primary while holding with Secondary it stays in the Secondary hand.
-Regrab with the Primary isnt possible and upon full drop the pickable cant be picked up again.


2b. *Swapping while holding with Primary to Primary is not possible, right now instead of swap from left to right ur other hand snaps instantly to the secondary. 
- i do get that this is due the setting of the pickables setting - second hand grab set to ignore but this is not what we want for a two handed item to behave.
- it should allow for swap from left to right as well as secondary

thats for now, some more testing will be done from my pov but 2a and 2b are sort of related to each other and really need to be resolved to make this a valid solution

BastiaanOlij (approved):
Amazing!

assets/3dmodelscc0/models/scenes/sniper_rifle.tscn
```diff
@@ -0,0 +1,118 @@
...
+closed_pose = ExtResource("3_np0oa")
+
+[sub_resource type="Resource" id="Resource_6toxn"]
+script = ExtResource("4_mworp")
+open_pose = ExtResource("7_ll2st")
+closed_pose = ExtResource("7_ll2st")
+
+[node name="SniperRifle" instance=ExtResource("1_lpv7q")]
+ranged_grab_method = 2
+second_hand_grab = 3
```
BastiaanOlij, line 42: There doesn't seem to be an  enum for `second_hand_grab = 3`, is this a left over from testing?

Malcolmnixon: Yes - there was an earlier attempt at simplifying the two handed grab approach into "grab" vs "aim" but the weighting system just provided better flexibility. This hold-over has now been fixed.

Malcolmnixon:
I debugged two-handed grabbing with the collision hands on the godot-xr-template project (https://github.com/GodotVR/godot-xr-template/pull/26).

Malcolmnixon:
> alright here it goes:
> 
> 1. *Upon Drop the positioning should be smoothed out/ lerped as it is a bit distracting and feels more like a springjoint.
> 
> * purely a visual feedback thing
> 
> 2a. *Drop while holding with Secondary should not make/force Secondary to be primary in position.
> 
> * On that note, when setting the Secondary to be a General one, upon drop of Primary while holding with Secondary it stays in the Secondary hand.
>   -Regrab with the Primary isnt possible and upon full drop the pickable cant be picked up again.
> 
> 2b. *Swapping while holding with Primary to Primary is not possible, right now instead of swap from left to right ur other hand snaps instantly to the secondary.
> 
> * i do get that this is due the setting of the pickables setting - second hand grab set to ignore but this is not what we want for a two handed item to behave.
> * it should allow for swap from left to right as well as secondary
> 
> thats for now, some more testing will be done from my pov but 2a and 2b are sort of related to each other and really need to be resolved to make this a valid solution

I believe I've handled the weird two-handed swapping issues.

DigitalN8m4r3:
> I believe I've handled the weird two-handed swapping issues.

lgtm, completes two handed

teddybear082:
This worked really well to me, only "issue" was that if I held two handed and purposely let go of primary hand and then not at the exact same time but immediately after before weapon had fully gotten to second hand, released second hand, sniper would fly off with unnatural force.  It may be the body thing Bastiaan mentions above but it didn't seem to happen when just naturally letting go of the gun for me, only with this specific combination.

This is a real step forward to let developers create that last percentage of games that previously weren't "ready to go" with xr tools.  Congrats!

PR #564, opened 2023-11-08
use get_normal() method instead of accessing property directly

duffrecords:
In  Godot 4, the `KinematicCollision3D` class has a `get_normal()` method instead of the `normal` property from its Godot 3.x counterpart, `KinematicCollision`.


Malcolmnixon (approved):
Good catch

PR #565, opened 2023-11-19
Fixed stream of editor errors with hand physics bones in the editor

Malcolmnixon:
The XRToolsHandPhysicsBone script needs to be a tool script for editor checks, but it failed to bypass _ready() and _physics_process() behavior in the editor so a stream of errors is produced.


TylerPottsDev:
This updated script fixed the issue for me, it no longer producing a stream of errors.

BastiaanOlij (approved):
lgtm!!

Issue #566, opened 2023-11-19
expose trigger on FunctionPickup for GrabPoints


DigitalN8m4r3:
Issue:
Pickable Objects with interactable handle part, when using twohanded, it is almost impossible to grab the handle, due to the grabpoints of the primary and secondary being simply in the way and overruling.

after some talk with @Malcolmnixon , we were able to come up with a limited solution for handguns and sorts,
this solution involves using an angular difference in the grabpoint. Still even with this, it only works for a limited type of pickable with interactable handles and it is prone to error due to its limitation of being dependant upon the axis and if u try to grab from another angle, it fails and grabs instead the secondary or swaps hand

here is a video of this

https://github.com/GodotVR/godot-xr-tools/assets/56046022/482c2f2d-5c5f-467a-9cc9-56b46a52b0bc

and the edited grab_point.gd - provided by @Malcolmnixon 
[grab_point.zip](https://github.com/GodotVR/godot-xr-tools/files/13403082/grab_point.zip)

Solution:
as mentioned, the above solution is pretty limited and causes more confusion and is not reliable enough, therefore the way to go
is to expose the trigger on the function pickup, so that handles can be picked up by pressing a trigger instead of grip.
this extra setting should look for the grabpoints isTrigger enabled setting and if it is, grab it when pressing the trigger

add Trigger to Function Pickup, check on grabpoint if trigger is prefered/enabled and if trigger pressed, snap hand to the handle.

this would ensure that we can have interactable handles on pickable items working.

DigitalN8m4r3:
after some more thought, think i got at least a less destructive solution...

this one involves the edit of the XRToolsFunctionPoseDetector and the XRToolsHandPoseArea

i have attached edited files, in a nutshell i have added an export option to connect to desired grabpoints, those grabpoints will then be disabled upon entering the handpose area and enabled upon leaving it.

Since we disable the secondary grabpoints upon entering the handpose area, we can now without issues grab the handle.
Good practice is obviously to not have the handpose area be around the primary grabpoints as this would obviously disable the secondary from the get go.

[grab_handle_solution.zip](https://github.com/GodotVR/godot-xr-tools/files/13404685/grab_handle_solution.zip)

PR #567, opened 2023-11-19
updated handpose detector and handpose area to disable grabpoints

DigitalN8m4r3:
edited the handpose detector and handpose area scripts to contain an export var for grabpoints, the user can specify grabpoints which are supposed to be disabled upon entering the area and re-enabled upon exit of the area

added firearm_bolt.gd and firearm_slide.gd
as an demonstrative addition, this pr adds to the sniper rifle a interactable handle - bolt to demonstrate visualy how the pr is supposed to be used and how it is working

Demo Video:

https://github.com/GodotVR/godot-xr-tools/assets/56046022/9ad74ebf-34fa-4772-84ce-ace335d4c5fc

this PR addresses
#566


Malcolmnixon (approved):
Looks and feels good when tested with the rifle.

Issue #568, opened 2023-11-25
Incorrect use of legacy extents property when setting screen shape


Malcolmnixon:
The BoxShape3D.extents has been replaced with BoxShape3D.size

https://github.com/GodotVR/godot-xr-tools/blob/b5d7fc1f2729b5dd122f1f6236148bc8f65a6fe4/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd#L322C44-L322C44

Malcolmnixon:
The legacy "extents" property is still supported in Godot 4.x for backwards compatibility inside an `#ifdef DISABLE_DEPRECATED` block; so it's probably a good idea to upgrade before the deprecated features are disabled.

Issue #569, opened 2023-11-25
Review changes to StandardMaterial3D properties


Malcolmnixon:
The `flags_transparent` boolean property was removed in Godot 4 and replaced with `transparency` enum. 

https://github.com/GodotVR/godot-xr-tools/blob/b5d7fc1f2729b5dd122f1f6236148bc8f65a6fe4/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd#L440C21-L440C21

Malcolmnixon:
The old `flags_transparent` is deprecated but still supported inside a `#ifndef DISABLE_DEPRECATED` block. We should fix it before the deprecated feature is removed.

PR #571, opened 2023-11-25
Upgrade use of legacy properties in the Viewport2Din3D script

Malcolmnixon:
This PR fixes issues #568 and #569 by:
- Switching from BaseMaterial3D.flags_transparent to BaseMaterial3D.transparency
- Switching from BoxShape3D.extents to BoxShape3D.size

### DIFF
--- a/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
+++ b/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
@@ -319,10 +319,10 @@ func set_filter(new_filter: bool) -> void:
 func _update_screen_size() -> void:
 	$Screen.mesh.size = screen_size
 	$StaticBody3D.screen_size = screen_size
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
When did this change? Is this something we forgot between Godot 3 and 4? Or something that was changed in 4.2?

Malcolmnixon:
> When did this change? Is this something we forgot between Godot 3 and 4? Or something that was changed in 4.2?

I found out when porting to TiltFiveTools C# that the properties weren't showing up in the intellisense popup. I checked back and found they have officially been deprecated since 4.0, but the resources are implemented with temporary hacks (custom generic set function with string matching) to handle the legacy properties. The hacks have ifdefs around them indicating they will be removed at some point soon.

BastiaanOlij:
If this was an old change in Godot 4.0 and the new properties have been available since that release, I'd say we should merge this.

Issue #578, opened 2023-12-07
Materials should be stored in "tres" files


Malcolmnixon:
Some of the materials in godot-xr-tools are stored in ".material" files which are binary files (actually .res files with a different extension). These files are opaque to source control and cannot be analyzed using diffing tools. As such these files should be converted to ".tres" files instead.

PR #579, opened 2023-12-07
Convert addon ".material" files to ".tres" files

Malcolmnixon:
This PR resolves issue #578  by converting the addon ".material" binary files to ".tres" text files to better handle source control and reviewing changes.


DigitalN8m4r3:
lgtm, no issues found

BastiaanOlij (approved):
I think we can get away with putting this into 4.3.1, definitely makes it easier to maintain.

One thing that is important, but I think this should always be the case, is that textures should always be separate from the material. Embedded textures in tres files become HUGE. This can happen depending on the import format and settings being used.

Issue #580, opened 2023-12-09
Jumping on sloping ground is broken


Malcolmnixon:
PR https://github.com/GodotVR/godot-xr-tools/pull/519 broke jumping while running on a sloped ground. Specifically the following test is broken:
https://github.com/GodotVR/godot-xr-tools/blob/c5c74338f88d393ee0724becf206f7d8dde744e0/addons/godot-xr-tools/player/player_body.gd#L350-L353
It calculates the players velocity (relative to the ground) and attempts to ascertain if there is a significant vertical component to it indicating a jump or landing is in effect.

The problem occurs when the player is running up or down a slope which involves a vertical velocity with regards to the players "up" vector. Instead the test should be against the grounds normal vector - asking whether there is a vertical velocity moving towards or away from the ground surface.

PR #581, opened 2023-12-09
Fix jumping on slopes

Malcolmnixon:
This PR fixes issue #580 by checking for vertical velocity against the current ground plane rather than the players up-vector. Essentially a player "running on the ground" should have a velocity vector with components only on the ground-surface plane, and no component aligned with the normal of the ground-plane.


DigitalN8m4r3:
seems to fix #580 
lgtm

BastiaanOlij (approved):
lgtm

PR #586, opened 2023-12-11
Automatically update hand poses in the editor

Malcolmnixon:
This PR updates the editor-preview hand-pose when changing the pose resource. As such the developer does not have to toggle between open/closed preview to see the updated hand pose.


BastiaanOlij (approved):
lgtm

PR #587, opened 2023-12-12
Improve the threaded staging for improved scene-switching performance.

Malcolmnixon:
This PR start the threaded loading of the new scene earlier in the scene-switch process, so the loading is happening at the same time the old scene is being faded out. If the new scene finishes loading before the fade is complete (and the user does not want a continue prompt) then the loading screen is bypassed and the new scene is immediately loaded and faded in.


BastiaanOlij (approved):
lgtm

PR #608, opened 2023-12-21
Fix world-grab movement

Malcolmnixon:
Fix world-grab movement to work with two-handed pickup changes.
Fix world-grab default area collision settings.
Update versions.md to include recent fixes.


BastiaanOlij (approved):
lgtm

PR #615, opened 2024-01-18
React to a recenter request by the headset

BastiaanOlij:
Still working on this but adding the logic into our `start_xr` script so we also react on a recenter request.

At this point in time this results in the player being positioned on the XROrigin point. That may be the right solution in some scenarios but not all. So will need to play a bit more with this.

### DIFF
--- a/addons/godot-xr-tools/xr/start_xr.gd
+++ b/addons/godot-xr-tools/xr/start_xr.gd
@@ -26,6 +26,13 @@ signal xr_started
 signal xr_ended
 
 
+## This signal is emitted when the user has requested to recenter their view.
+## The default logic already ensures that the player is now centered on
+## the XROrigin facing the -z direction.
+## You can further react to this by placing the XROrigin where the player
+## should be positioned.
+signal xr_recentered
+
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
> nice this rounds things up, indeed recenter on xrorigin might fit for the current Version, since malcolm is working on the character centric solution, i dont see why the above shouldnt be merged prior. It would in essence complement the current version before the big breaking changes jump.

The thing that is still missing is that this jumps the player back to the origin point, both when this is desired, and when it is not.
So needs a little bit more experimentation to get this right.

DigitalN8m4r3:
> > nice this rounds things up, indeed recenter on xrorigin might fit for the current Version, since malcolm is working on the character centric solution, i dont see why the above shouldnt be merged prior. It would in essence complement the current version before the big breaking changes jump.
> 
> The thing that is still missing is that this jumps the player back to the origin point, both when this is desired, and when it is not. So needs a little bit more experimentation to get this right.

indeed, right..  i got a somewhat special case to test the above.
a hide system which sets the current origin to a new one. I wonder how well this will work when using the recenter and changing from seated to standing and back

PR #624, opened 2024-03-18
Add check/setting for foveated rendering

BastiaanOlij:
This PR adds a bit of code to our start xr script that:
- in Vulkan sets our `vrs_mode` to VRS_XR which, when supported, enables foveated rendering
- in compatibility, checks if our project setting has foveated rendering enabled and warns the user if this is not the case.

Note that this will bring the minimum Godot version to 4.2.x


BastiaanOlij:
Needs rebase and then can be merged.

PR #627, opened 2024-04-01
Hands: fix open and close pose script

### LINKED ISSUE #589: There is no 'add_animation' function in AnimationPlayer any all
In res://addons/godot-xr-tools/hands/hand.gd. 
image

Here try to add animation to AnimationPlayer, but this func is deprecated in 4.2. 
Instead, you have to get animation library first and then add animation to the library. 
The same to add "close pose" after this.

esodan:
In order to add an animation, AnimationPlayer
should get its AnimationLibrary with empty name "".

This should fix #589


Malcolmnixon (approved):
This looks good to me.

BastiaanOlij (approved):
lgtm

PR #630, opened 2024-04-01
Fix issue629

esodan:
This should be merged only if #627 is fixed first, as this depends on it

### DIFF
--- a/addons/godot-xr-tools/staging/staging.gd
+++ b/addons/godot-xr-tools/staging/staging.gd
@@ -68,13 +68,30 @@ signal xr_ended
 		update_configuration_warnings()
 
 ## If true, the player is prompted to continue
+## so no change to main scene, until the user
+## interacts with the button to continue.
+## If false, the transition to main scene
+## is automatic
 @export var prompt_for_continue : bool = true
 
-## Scene container file as a Node3D used as
+## Scene container file as a [Node3D] used as
 ## the parent of the current scene.
 @export var scene_container : Node3D:
 	set(value):
 		scene_container = value
+		update_configuration_warnings()
+
+## Fade is a [MeshInstance3D] object using a [QuadMesh]. Set its
+## shader to [godot-xr-tools/staging/fade.gdshader].
+##
+## Holds the fade-alpha for scene transitions. See set_fade method
+##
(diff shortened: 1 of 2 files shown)


Malcolmnixon:
The hand change looks good, but if you want to do any changes to staging; could you do that in a separate PR.

addons/godot-xr-tools/staging/staging.gd
```diff
@@ -269,12 +284,12 @@ func load_scene(p_scene_path : String, user_data = null) -> void:
 ## unless hidden.
 func set_fade(p_value : float):
 	if p_value == 0.0:
-		$Fade.visible = false
+		fade.visible = false
```
BastiaanOlij, line 287: I think you lost a change here. The variable fade is not defined anywhere or set anywhere.

BastiaanOlij:
It would be good to have more of an explanation in your description on why you are changing this.

The staging system is meant to be used with the staging scene, which is a scene you can inherit if there is more that you want to add in. This just introduces more work for a user to set things up that should already be there.

So without knowing the justification for this change, it breaks the "keep things simple" rule.

esodan:
> It would be good to have more of an explanation in your description on why you are changing this.
> 
> The staging system is meant to be used with the staging scene, which is a scene you can inherit if there is more that you want to add in. This just introduces more work for a user to set things up that should already be there.
> 
> So without knowing the justification for this change, it breaks the "keep things simple" rule.

The comunity provides lot of tutorials, most of them helps you to "walk" for each step, adding by hand each node. This is good to know better how to make your own stuff.

After following that tutorials, make lot of mistakes and found this missing, underdocumented things like Fade. So if the user wants to do it by hand, this should be better to be added as a warning. Yes there are more steps here, but that is the price to do it by hand.

May is better to document, that the user should instantiate the staging scene, instead to create all step by step.

BastiaanOlij:
Discussing this further with @Malcolmnixon , we need to double check the hands fix and make sure that was applied.

As for the changes to staging, the `staging.gd` script is part of the `staging.tscn` file which contains all relevant nodes. This looks like an issue with using the script without the scene, which is something Godot allows while it shouldn't.

This needs better tutorials around the staging system but not code changes.

Closing this PR.

PR #631, opened 2024-04-02
Move fading to effect and add collision fading support.

Malcolmnixon:
This PR moves the fade logic to a common effect accessible by all nodes - including with the ability to fade to colors.
Additionally it adds collision-based fading to prevent looking through walls.

### DIFF
--- a/addons/godot-xr-tools/effects/fade.gd
+++ b/addons/godot-xr-tools/effects/fade.gd
@@ -1,0 +1,79 @@
+@tool
+class_name XRToolsFade
+extends Node3D
+
+
+## XR Tools Fade Script
+##
+## This script manages fading the view.
+
+
+# Dictionary of fade requests
+var _faders : Dictionary = {}
+
+# Fade update flag
+var _update : bool = false
+
+# Fade mesh
+var _mesh : MeshInstance3D
+
+# Fade shader material
+var _material : ShaderMaterial
+
+
+# Add support for is_xr_class on XRTools classes
+func is_xr_class(name : String) -> bool:
+	return name == "XRToolsFade"
+
+
+# Called when the fade node is ready
+func _ready() -> void:
+	# Add to the fade_mesh group - in the future this should be replaced with
+	# static instances.
+	add_to_group("fade_mesh")
+
+	# Get the mesh and material
+	_mesh = $FadeMesh
+	_material = _mesh.get_surface_override_material(0)
+
+
+# Called every frame. 'delta' is the elapsed time since the previous frame.
+func _process(_delta : float) -> void:
+	# Skip if nothing to update
+	if not _update:
+		return
+
+	# Calculate the cumulative shade color
+	var fade := Color(1, 1, 1, 0)
+	var show := false
+	for whom in _faders:
+		var color := _faders[whom] as Color
+		fade = fade.blend(color)
+		show = true
+
+	# Set the shader and show if necessary
+	_material.set_shader_parameter("albedo", fade)
+	_mesh.visible = show
+	_update = false
+
+
+# Set the fade level
+func set_fade_level(p_whom : Variant, p_color : Color) -> void:
+	# Test if fading is needed
+	if p_color.a > 0:
+		# Set the fade level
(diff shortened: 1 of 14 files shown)


DigitalN8m4r3:
lgtm, thanks again malcolm for working this out!

Malcolmnixon:
Added a basic damage area to the basic movement scene.

https://github.com/GodotVR/godot-xr-tools/assets/1863707/74977229-2e61-495f-9bbc-7100c101e8e5

addons/godot-xr-tools/effects/fade.gdshader
```diff
@@ -1,13 +1,13 @@
 shader_type spatial;
 render_mode depth_test_disabled, skip_vertex_transform, unshaded, cull_disabled;
 
-uniform float alpha = 0.0;
+uniform vec4 albedo : source_color;
 
 void vertex() {
 	POSITION = vec4(VERTEX.x, -VERTEX.y, 0.0, 1.0);
```
BastiaanOlij, line 7: Something we need to change here that I didn't realise until later is that this needs to be `POSITION = vec4(VERTEX.x, -VERTEX.y, -1.0, 1.0);` when using OpenGL (compatibility). The code as it stands right now is correct for Vulkan.

Once the reverse-Z PR is merged in 4.3, we need to change this again to `POSITION = vec4(VERTEX.x, -VERTEX.y, 1.0, 1.0);`

addons/godot-xr-tools/effects/fade.gdshader
```diff
@@ -1,13 +1,13 @@
...
 }
 
 void fragment() {
-	ALBEDO = vec3(0.0, 0.0, 0.0);
-	ALPHA = alpha;
+	ALBEDO = albedo.rgb;
+	ALPHA = albedo.a;
 }
```
BastiaanOlij, line 13: Nitpicking, needs newline at the end

BastiaanOlij (approved):
Looks good, cool addition

addons/godot-xr-tools/effects/fade.gdshader
Malcolmnixon: This shader works on the current options (tested on PC, not native):
- Godot 4.1.3 Vulkan & OpenGL
- Godot 4.2.1 Vulkan & OpenGL
- Godot 4.3-dev5 Vulkan & OpenGL

Back in 4.0 we were looking for an option that would satisfy both OpenGL and Vulkan, and I thought the ``cull_disabled` was all we needed.

addons/godot-xr-tools/effects/fade.gdshader
Malcolmnixon: Good catch - fixed.

addons/godot-xr-tools/effects/fade.gdshader
Malcolmnixon: What would we need to do if we want this shader to work in Vulkan & OpenGL on Godot 4.1 - 4.3?

In the Reverse Z Progress Report announcement there didn't seem to be anything about how to handle forwards/backwards compatibility.

addons/godot-xr-tools/effects/fade.gd
```diff
@@ -0,0 +1,79 @@
...
+	var fade := Color(1, 1, 1, 0)
+	var show := false
+	for whom in _faders:
+		var color := _faders[whom] as Color
+		fade = fade.blend(color)
+		show = true
+
+	# Set the shader and show if necessary
+	_material.set_shader_parameter("albedo", fade)
+	_mesh.visible = show
```
BastiaanOlij, line 56: You'll want to add an extra check here if alpha = 0. If the fade is invisible but you don't make the node invisible, you're drawing a full screen rect that is blending with a fully transparent colour. That is unnecessary overhead.

BastiaanOlij: Never mind, just noticed the alpha check below :P

BastiaanOlij (approved):
LGTM

PR #637, opened 2024-05-24
Viewport2Din3D: Forward properties of supplied scene object

NetroScript:
This PR adjusts the `viewport_2d_in_3d.gd` script to allow forwarding / proxying of the scene provided as PackedScene. 

Currently, to assign properties to the scene shown in the Viewport, it would be necessary to enable editable children, instance the scene as child, and assign properties to that node.

This however adds noise to the SceneTree, which could be reduced if these properties could be directly editable without needing the children in the tree.

-------------------------

The script works by reading all properties of the script of the packaged scene, and forwarding the property accesses (reading and writing) to the object itself it it exists. On scene load before it is instantiated at least once, a proxy object is used, where then the values are applied to the instanced scene as soon as it loads.

To get properties `Script.get_script_property_list()` is currently employed, this means only custom defined exports of the script are displayed (not inherited properties).

#### Example Image:
Screenshot_20240524_140514

#### Thoughts

It might be worth a consideration if one wants to extend this to the current class if no script is attached to the scene (using `ClassDB.class_get_property_list(class, true)`) or if even all editable properties of the underlying object should be exposed. (Although then it would probably better to create a dummy Resource which contains all of those properties to have less clutter and a clear separation between the current node, and all properties of the embedded node). 

#### Potential Issues

* Properties with the same name

Currently if a property has the same name in the provided scene and the Viewport2Din3D script, the property assignment will always be done to the provided scene, and it won't be possible to access this value for the Viewport2Din3D script anymore.

* Relative values

As these values are applied to the parent (which then forwards it to the scene) for the Editor, all relative things might need special handling. This PR already handles this for `NodePath` by adjusting the value when getting / setting by prepending / removing `../../`.

### DIFF
--- a/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
+++ b/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
@@ -110,6 +110,11 @@ var filter : bool = true: set = set_filter
 
 var is_ready : bool = false
 var scene_node : Node
+var scene_properties_keys: PackedStringArray = []
+var scene_properties : Array[Dictionary] = []
+# Needed to apply custom properties of the scene before it is instanced, as these are set on ready, 
+# But at this point in time the scene is not instanced yet
+var scene_proxy_configuration: Dictionary = {}
 var viewport_texture : ViewportTexture
 var time_since_last_update : float = 0.0
 var _screen_material : StandardMaterial3D
@@ -139,8 +144,7 @@ func _get_property_list() -> Array[Dictionary]:
 	var show_unshaded := not material
 	var show_filter := not material
 
-	# Return extra properties
-	return [
+	var extra_properties : Array[Dictionary] = [
 		{
 			name = "Rendering",
 			type = TYPE_NIL,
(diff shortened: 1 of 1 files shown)


Malcolmnixon:
This is looking quite good; but there are some gdlint issues being reported in the new code:
```
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:221: Error: Max allowed line length (100) exceeded (max-line-length)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:274: Error: Max allowed line length (100) exceeded (max-line-length)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:115: Error: Trailing whitespace(s) (trailing-whitespace)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:170: Error: Trailing whitespace(s) (trailing-whitespace)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:180: Error: Trailing whitespace(s) (trailing-whitespace)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:202: Error: Trailing whitespace(s) (trailing-whitespace)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:215: Error: Trailing whitespace(s) (trailing-whitespace)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:264: Error: Trailing whitespace(s) (trailing-whitespace)
.\godot-xr-tools\objects\viewport_2d_in_3d.gd:270: Error: Private method "_get_property_list" has been called (private-method-call)
```

Most of them are simple formatting; but you may want to investigate calling the public `get_property_list()` method exposed by all godot Objects rather than the private `_get_property_list()`.

NetroScript:
Fixed the linting issues.

Malcolmnixon (approved):
This looks good and works well in my testing. Good job!

PR #640, opened 2024-06-04
Fix slope-sliding by applying ground-friction at the correct time

### LINKED ISSUE #639: Incorrect slope behavior
Something reported by HrenEco, he prepared the MRP.

I'm not sure if we just never noticed this or whether there is some difference I haven't discovered yet compared to some of the slope demos in XR tools, but a player giving no input while on a slope, slowly moves down the slope.

Simply start the MRP, and look a the blocks. you'll see them slowly move past as you slide down the slope.

I did some testing with a simple physics scene and things worked correctly there. My best guess is that this is a result due to us splitting moving the player down by gravity, and moving the player on user input. I think move and slide can't accurately perform its slope check in this scenario and our movement is caused by simple rounding error that gets compounded. We may need to rethink this.

MRP:
[godot-sliding.zip](https://github.com/user-attachments/files/15524981/godot-sliding.zip)

### LINKED ISSUE #642: Ground-control prevents jumping over walls
When ground-control is set for "Always", the player cannot jump over walls even if they're high above the wall.

Malcolmnixon:
This PR fixes #639 by applying the ground-friction at the correct time in the movement calculations.
This PR also fixes #642 by only applying the steep-slope logic if the player is on the ground.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,8 @@
 # 4.3.2
 - Move fade logic into effect
 - Added collision fade support
+- Added fix for slowly sliding on slopes
+- Added fix for ground-control preventing jumping over objects
 
 # 4.3.1
 - Fix saving project when using plugin-tools to set physics layers or enable OpenXR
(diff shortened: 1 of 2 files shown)


HrenEgo:
ground_control = Always (If the player does not move after jumping while falling, the player falls slowly to the ground. If the player moves in any direction during the jump, all is well.)
ground_control = On Ground (It's okay

Malcolmnixon:
Updated how friction is applied so it's only applied horizontally in the reference frame of the ground.

BastiaanOlij (approved):
LGTM

HrenEgo:
Everything's great. No problems. Thank you for your work!!!

PR #643, opened 2024-06-11
Change wall-walking to use the Godot 4.x version of move_and_collide

Malcolmnixon:
This PR corrects the wall-walking move_and_collide to use the Godot 4.x function parameters rather than the 3.x parameters. 

Specifically the code is using move_and_collide to get a KinematicCollision for the object the player is standing on.

It was doing this by passing the following parameters which are valid for Godot 3.x:
1. Movement vector
2. true (default)
3. true (default)
4. true for test_only (to test without performing the move)

However for Godot 4.x the test_only moved to the second parameter. The third and fourth parameters are the float safety_margin and the boolean recovery_as_collision. Providing true to these results in unreliable KinematicCollision data.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -3,6 +3,7 @@
 - Added collision fade support
 - Added fix for slowly sliding on slopes
 - Added fix for ground-control preventing jumping over objects
+- Fix unreliable wall-walking collision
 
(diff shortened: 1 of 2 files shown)


BastiaanOlij (approved):
lgtm

PR #645, opened 2024-06-19
Updating to Godot 4.2

BastiaanOlij:
This PR bumps the minimum version of Godot up to 4.2.

This includes updating of mesh data to the 4.2 format.

I've also cleaned up the teleporter scene a bit so it uses the GLB directly and saves the convex collision shapes as `.shape` files.
Similar changes should be made elsewhere as ArrayMesh and Convex shape data is stored very inefficiently in `.tscn` files.


README.md
```diff
@@ -12,11 +12,26 @@ This repository contains a number of support files and support scenes that can b
...
+|  master   | Current development branch    |  Godot 4.2+     |
+|  4.3.x    | Godot XR Tools 4.3.x branch   |  Godot 4.1+     |
 |   3.x     | Godot 3.x development branch  |  Godot 3.5+     |
 
+Version compatibility:
+
+| Godot XR Tools | Minimum Godot version |
+|----------------|-----------------------|
+| 4.4.0          | 4.3                   |
+| 4.3.2          | 4.1                   |
```
Malcolmnixon, line 27: Is 4.4.0 going to support Godot 4.3 or 4.2? Additionally perhaps we just use `x` in place of the patch version. Something like:

| Godot XR Tools | Minimum Godot version |
|----------------|-----------------------|
| 4.4.x          | 4.2                   |
| 4.3.x          | 4.1                   |
| 4.2.x          | 4.1                   |
| 4.1.x          | 4.0                   |
| 4.0.x          | 4.0                   |

Malcolmnixon:
I'm not sure how far we want to go with the teleporter cleanup in this PR; but I found that if I correctly defined material entries in the blend/glb files, then I could attach Godot materials in the GLB importer, and have it construct the colliders (simple convex) to produce all visual and physics artifacts via import rather than post-processing in the scene.

The following is the updated blend/glb file with the material slots named so the GLB importer can correctly assign to them. [teleport_glb_blend.zip](https://github.com/user-attachments/files/15930616/teleport_glb_blend.zip)

README.md
BastiaanOlij: I think that we should stick to 4.2, it's mostly around the mesh conversions and such so..

BastiaanOlij:
> I'm not sure how far we want to go with the teleporter cleanup in this PR; but I found that if I correctly defined material entries in the blend/glb files, then I could attach Godot materials in the GLB importer, and have it construct the colliders (simple convex) to produce all visual and physics artifacts via import rather than post-processing in the scene.
> 
> The following is the updated blend/glb file with the material slots named so the GLB importer can correctly assign to them. [teleport_glb_blend.zip](https://github.com/user-attachments/files/15930616/teleport_glb_blend.zip)

Yeah I went more down a rabbit hole there than I was expecting. I think its best to just merge and then do a follow up PR to improve?

BastiaanOlij:
I've also changed the default renderer to Compatibility and configured foveated rendering.

Malcolmnixon (approved):
Looks good now!

Issue #646, opened 2024-06-19
New, incorrect behavior of Viewport2Din3D. XR-Tools 4.3.2, Godot 4.3.beta1


Eisenspalter:
New, incorrect behavior of Viewport2Din3D: After setting the Properties menu and reloading the scene, the menu script is automatically set to Viewport2Din3D. This then leads to further errors. 
![image](https://github.com/GodotVR/godot-xr-tools/assets/4734410/ef4e678a-3cec-4054-a549-e4830201213c)

Malcolmnixon:
@Eisenspalter Could you try out the change in the PR and confirm it fixes your issue.

PR #647, opened 2024-06-20
Fix properties forwarding to only forward customized properties.

Malcolmnixon:
This PR fixes bug #646 in property-forwarding for Viewport2Din3D introduced by gdlint cleanup whereby **ALL** properties of the sub-scene (including its script) were copied to the Viewport2Din3D node.


BastiaanOlij (approved):
LGTM

Eisenspalter:
Works now

PR #648, opened 2024-06-25
Simplify teleporter object

Malcolmnixon:
This PR simplifies the teleporter by:
* Updating the blend/glb to add missing materials.
* Update the glb import to apply the desired materials at import-phase
* Simplify teleporter disabling by setting the process mode to PROCESS_MODE_DISABLED

The PROCESS_MODE_DISABLED trick works because the StaticBody3D nodes have their disable mode set to "Remove".

Note: configurable materials (the transporter effect and title) still need to be set in "Surface Material Override" as this lets them be marked as local-to-scene so they can have individual shader properties.


BastiaanOlij (approved):
lgtm

PR #649, opened 2024-06-30
Clean up StartXR script

Malcolmnixon:
This PR improves the start_xr script by:
* Exposing target XR frame rate
* Removing obsolete auto_initialize
* Using modern signal.emit() syntax
* Setting physics rate on WebXR
* Enabling WebXR hand-tracking if enabled in project settings
* Enabled WebXR passthrough support

Note: This is in preparation for improving the passthrough blend-mode support as a future PR.

### DIFF
--- a/addons/godot-xr-tools/xr/start_xr.gd
+++ b/addons/godot-xr-tools/xr/start_xr.gd
@@ -9,8 +9,7 @@ extends Node
 ## the initialization of the interface as well as reporting when the user
 ## starts and ends the VR session.
 ##
-## For OpenXR this class also supports passthrough on compatible devices such
-## as the Meta Quest 1 and 2.
+## For OpenXR this class also supports passthrough on compatible devices.
 
 
 ## This signal is emitted when XR becomes active. For OpenXR this corresponds
@@ -25,9 +24,12 @@ signal xr_started
 ## signal.
 signal xr_ended
 
+## This signal is emitted if XR fails to initialize.
+signal xr_failed_to_initialize
 
-## If true, the XR interface is automatically initialized
-@export var auto_initialize : bool = true
+
+## Optional viewport to control
+@export var viewport : Viewport
 
 ## Adjusts the pixel density on the rendering target
 @export var render_target_size_multiplier : float = 1.0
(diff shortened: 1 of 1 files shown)


addons/godot-xr-tools/xr/start_xr.gd
```diff
@@ -91,9 +90,7 @@ func _setup_for_openxr() -> bool:
 	print("OpenXR: Configuring interface")
 
 	# Set the render target size multiplier - must be done befor initializing interface
```
BastiaanOlij, line 92: I would also remove the `must be done before initializing interface` bit as initialising happens before any of our code is run. In Vulkan this actually fully works now. In OpenGL there is still an issue with releasing things because we're lacking dependency checking but the buffers aren't allocated until the first frame is rendered after `use_xr` is set.

addons/godot-xr-tools/xr/start_xr.gd
```diff
@@ -48,13 +44,16 @@ var xr_interface : XRInterface
...
+var xr_frame_rate : float = 0
+
+# Is a WebXR is_session_supported query running
+var _webxr_session_query : bool = false
 
 
 # Handle auto-initialization when ready
 func _ready() -> void:
-	if !Engine.is_editor_hint() and auto_initialize:
+	if !Engine.is_editor_hint():
```
BastiaanOlij, line 56: Seeing we decided to remove `auto_initialize`, we should rename our `initialize` function to `_initialize` to indicate it's private.
We should also add an `xr_failed_to_initialize` signal that we emit when `_initialize` fails so the user can react to this scenario, either by setting the game up for non-VR or to exit the app.
Similarly maybe we should also add an `xr_initialized` signal on successful initialisation for the same reason though that feels like overkill.

BastiaanOlij (approved):
Other than some nitpicking, looks good :)

addons/godot-xr-tools/xr/start_xr.gd
Malcolmnixon: I've added the signal and I trigger it if:
* OpenXR fails to initialize
* WebXR fails to initialize
* No XRInterface is found

addons/godot-xr-tools/xr/start_xr.gd
Malcolmnixon: I've removed the comment.

addons/godot-xr-tools/xr/start_xr.gd
Malcolmnixon: I've also renamed initialize to _initialize

BastiaanOlij:
Looks good!

PR #653, opened 2024-07-05
Fix custom hand poses triggering legacy remove_animation call

### LINKED ISSUE #650: Hand pose animation override not working with Godot 4.2
Hand animation override is not working for me when changing the hand pose by GrabPointHand. The game logs error
```
USER SCRIPT ERROR: Invalid call. Nonexistent function 'remove_animation' in base 'AnimationPlayer'.
    at: XRToolsHand._update_pose (res://addons/godot-xr-tools/hands/hand.gd:346)
```

Calls to `_animation_player.remove_animation(open_name)` (and `remove_animation(close_name)` few lines later) causes the trouble, replacing it by `_animation_player.get_animation_library("").remove_animation(open_name)` seems to fix the problem. 

My knowledge of the framework is pretty low though, I'm not really confident the fix wouldn't break something else.

Malcolmnixon:
This PR fixes #650 by using the players "" global animation library for both adding and removing custom hand poses.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,7 @@
 # 4.4.0
 - Minimum Godot version changed to 4.2
 - Add pickable action_released signal
+- Fix custom hand poses calling legacy remove_animation
 
 # 4.3.3
 - Fix Viewport2Din3D property forwarding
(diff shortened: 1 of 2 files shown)


BastiaanOlij (approved):
lgtm

Issue #652, opened 2024-07-04
Feature Request: action_released (signal) XRToolsPickable


gspinz:
Hi,

It would be useful to tell when the action button has been released on the pickable objects.

Thanks for your hard work on this plugin.

Malcolmnixon:
The concept of the "action" button is a legacy idea from the earlier days of Godot-XR-Tools and is somewhat retired at this point.

Consider that a function-pickup can only have one "action" button set - usually the trigger, but when you're holding different objects they would want to react to different user inputs. For example:
- A flash-light may want to react to an AX or BY button
- A gun with laser-scope would want two actions (trigger=fire, and AX=laser-toggle)

In the Godot-XR-Tools-Demo project `scenes\pickable_demo\objects\grab_cube.gd` shows how to have an object subscribe to controller events when picked up, and unsubscribe when dropped.

I may put together a PR just to make the legacy "action" button have both `action_pressed` and `action_released` for completeness, but instead we may revisit the concept for 4.4 and remove or replace it.

Malcolmnixon:
@gspinz Could you confirm the action_released behavior satisfies your needs, and if so close the issue.

gspinz:
Hi Malcolm,

Thanks for the response. Yes, action_released would be quite useful for those who are new to godot. 

Thank you for the XR Demo project link. It has been very helpful.

PR #654, opened 2024-07-05
Add signal when action button is released

Malcolmnixon:
This PR implements feature request #652, however due to the limitations of the single "action" button we may end up replacing/retiring this concept for either 4.4 or possibly the Godot-XR-Tools-2 rewrite.


BastiaanOlij (approved):
I think this is an ok way of doing it for now. I'm not even sure if in two handed mode we react on both triggers but I also think that is a real edge case.

PR #656, opened 2024-07-08
Add support for alternate hand-trackers

Malcolmnixon:
This PR allows hand-poses and grab-points to work with XRController3D nodes driven by alternate sources such as virtual controllers created from hand-trackers.


BastiaanOlij (approved):
lgtm

PR #658, opened 2024-07-13
Add end_xr support

Malcolmnixon:
This PR adds the end_xr() method to StartXR. This can be called directly, or can be triggered using a quit request routed through staging. The example wrist control-pad now includes a Quit button which triggers this staging-quit.

Note: Also added the is_xr_active() static method.

Note: As of Godot 4.3-beta3 this triggers a stack-overflow in WebXR, but the overflow seems to occur even when using the web-browser quit button.


BastiaanOlij (approved):
lgtm

PR #659, opened 2024-07-13
Add visibility_changed notifications to Viewport2Din3D hosted scenes

Malcolmnixon:
This PR triggers visibility_changed notifications to 2D scenes hosted in a Viewport2Din3D when the 3D visibility changes.

Additionally as requested, this PR also disable physics and viewport rendering when the Viewport2Din3D is invisible

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -6,6 +6,8 @@
 - Allow grab-points and poses to work with different types of hand trackers
 - Add end_xr support to StartXR
 - Fixed vignette shader
+- Add visibility_changed notifications to Viewport2Din3D hosted scenes
+- Invisible Viewport2Din3D now disable physics and viewport updates
 
 # 4.3.3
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
If a Viewport2Din3D is not visible we should:
- set the update mode to disabled to stop wasting rendering resources
- disable the collision shape
When it is visible:
- set the update mode depending on our throttle logic as it is right now
- enable the collision shape

Malcolmnixon:
> If a Viewport2Din3D is not visible we should:
> 
> * set the update mode to disabled to stop wasting rendering resources
> * disable the collision shape
> 
> When it is visible:
> 
> * set the update mode depending on our throttle logic as it is right now
> * enable the collision shape depending on user setting

Done

BastiaanOlij (approved):
lgtm

PR #660, opened 2024-07-19
Pass visibility changes on to the embedded 2d scene in viewport_2d_in_3d

DanielKinsman:
I had some logic in a 2D scene that was checking "hey am i visible" and it was happily reporting "yes" even when the viewport was hidden.

Though maybe this change will cause issues if you have multiple 2d viewports hanging in 3d space sharing the same `scene_node`, if that's even possible?


DanielKinsman:
Oops nevermind i see now that there was already a PR for this #659

Issue #662, opened 2024-07-20
Allow Gaze tracking for function pointer when hands not available.


RangerMauve:
At the moment the code assumes there are controllers present for pointers. It'd be nice if one could attach a ray to any 3d object like the camera.

If not in the FunctionPointer class, it'd be nice if some of the core raycasting logic was exposed somewhere reusable instead.

surreal6:
now #683 is the new PR for this feature

Malcolmnixon:
Can you confirm the solution from surreal6 meets your needs.

BastiaanOlij:
Assuming this is resolved with surreals additions.

PR #664, opened 2024-07-31
Gaze selector inside function_pointer   (accesibility feature)

surreal6:
This is a proof of concept to integrate a Gaze selector in function_pointer node, as an accesibility feature.

It works by adding a function pointer to the XRCamera.

may need to clean up some code, but i share it here for anyone to test it and give feedback

There is a problem with viewport2din3d, because the pointer detects the whole viewport as one element, and don't know if it is possible to detect individual items inside the viewport. I manage to solve this by limiting the pointer function to one press for each targeted element, so you need to look outside the viewport and back again inside to press a diferent element.

The release is always triggered after the press event

I added the function pointer to the XRCamera in the main scene and also in the pointer demo.

Also, there is a   **WIP/shead_selector_function_pointer.tscn** scene with a basic demo working with XRInteractableAreas

The idea before to merge this (if approved) is to create a scene in which a decision is made from among several options.

Other idea is to add the elements from the head selector scene into the pointer demo room.

this may be a solution for #662

https://github.com/user-attachments/assets/af0b2776-6fb6-406e-b80b-83385dd2cdb0

https://github.com/user-attachments/assets/3e6c713f-4e9b-4b8f-b80a-2e71fbfa87b3

https://github.com/user-attachments/assets/c85c91a9-4a7f-4e9e-aed5-66d8dbb76bb9


DigitalN8m4r3:
Lgtm, nice work there

Malcolmnixon:
Could you rebase to resolve the conflict - also what is the "WIP" folder - "Work in Progress"?

surreal6:
Hi, sorry for the delay fixin this. I rebased and now the gaze selector in main menu only works with the hmd off head. 

I have to look at latest commits to see what happens :P

In pointer scene works as expected

BastiaanOlij:
While still WIP it would be good to turn this into a draft PR.

Functionality is valuable but wonder if it makes more sense to make this a separate `gaze_pointer` scene/class you attach to the head instead of mixing it up with the current controller pointer logic.

surreal6:
> While still WIP it would be good to turn this into a draft PR.
> 
> Functionality is valuable but wonder if it makes more sense to make this a separate `gaze_pointer` scene/class you attach to the head instead of mixing it up with the current controller pointer logic.

It makes sense. In the next days I will make a new PR 

In my project i am exploring a diferent solution, i recreated the 'click on hold' feature in a traditional 2d UI, so i just need to use the pointer as is, allowing a camera parent and remoting the mouse, and without changing all the _process logic, and all the stuff detecting different ui elements is managed in the UI control nodes.

so i am thinking i will make the 'click on hold' feature optional in the new PR, to allow this other approach too.

surreal6:
I added a new pull request #683 for this functionality

PR #665, opened 2024-08-07
Fix things trying to access non-existent property on start xr

DanielKinsman:
`xr_active` was made private


DanielKinsman:
`is_xr_active()` is a static func now, should I also change the calls to use the class rather than the instance?

Malcolmnixon (approved):

Malcolmnixon:
This looks good to me. We're going to be doing some other improvements later with this - improving documentation and fixing a few spelling mistakes; but this works well for me in the main scene.

PR #666, opened 2024-08-07
enable actions on hand snap points

DanielKinsman:
Handling pickables like two handed guns where you only want the "trigger" hand to fire it [seems complex](https://github.com/GodotVR/godot-xr-tools/blob/master/assets/3dmodelscc0/models/sniper_rifle/firearm_trigger.gd#L58). I think it is simpler to add actions to the grab points instead of the pickable.

### DIFF
--- a/addons/godot-xr-tools/functions/function_pickup.gd
+++ b/addons/godot-xr-tools/functions/function_pickup.gd
@@ -413,13 +413,13 @@ func _pick_up_object(target: Node3D) -> void:
 func _on_button_pressed(p_button) -> void:
 	if p_button == action_button_action:
 		if is_instance_valid(picked_up_object) and picked_up_object.has_method("action"):
-			picked_up_object.action()
+			picked_up_object.action(_controller)
 
 
 func _on_button_released(p_button) -> void:
 	if p_button == action_button_action:
 		if is_instance_valid(picked_up_object) and picked_up_object.has_method("action_release"):
-			picked_up_object.action_release()
+			picked_up_object.action_release(_controller)
 
 
 func _on_grip_pressed() -> void:
--- a/addons/godot-xr-tools/functions/function_pickup.gd
+++ b/addons/godot-xr-tools/functions/function_pickup.gd
@@ -411,16 +411,20 @@ func _pick_up_object(target: Node3D) -> void:
 
 
 func _on_button_pressed(p_button) -> void:
-	if p_button == action_button_action:
-		if is_instance_valid(picked_up_object) and picked_up_object.has_method("action"):
+	if p_button == action_button_action and is_instance_valid(picked_up_object):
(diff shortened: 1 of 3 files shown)


DanielKinsman:
This is a quick fix which can perhaps be solved in a more holistic way, so happy to close it in favour of something more comprehensive. During the XR meeting someone mentioned just passing the controller events on down to the object directly, that would be nice. Perhaps something like a "pressed" event which tells the object:

* what input was pressed
* what controller pressed it
* where that controller is holding me

BastiaanOlij (changes requested):
Generally happy with this change except that it will break existing applications that implement the current action function.

The suggestions here is to change to keep both. So call `action` as is and leave the `action` implementation on `pickable.gd` as is currently.

Then call a new `controller_action` method that has the new code.

addons/godot-xr-tools/functions/function_pickup.gd
```diff
@@ -414,12 +414,14 @@ func _on_button_pressed(p_button) -> void:
 	if p_button == action_button_action:
 		if is_instance_valid(picked_up_object) and picked_up_object.has_method("action"):
 			picked_up_object.action()
+			picked_up_object.controller_action(_controller)
```
Malcolmnixon, line 417: The has_method check should be used to check for the new controller_action method as well. I would recommend something like:
```gdscript
	if p_button == action_button_action and is_instance_valid(picked_up_object):
		if picked_up_object.has_method("action"):
			picked_up_object.action()
		if picked_up_object.has_method("controller_action"):
			picked_up_object.controller_action(_controller)
```

addons/godot-xr-tools/functions/function_pickup.gd
```diff
@@ -414,12 +414,14 @@ func _on_button_pressed(p_button) -> void:
...
 
 
 func _on_button_released(p_button) -> void:
 	if p_button == action_button_action:
 		if is_instance_valid(picked_up_object) and picked_up_object.has_method("action_release"):
 			picked_up_object.action_release()
+			picked_up_object.controller_action_release(_controller)
```
Malcolmnixon, line 424: Same as the _on_button_pressed - this should check for the new method before calling it - something like:
```gdscript
	if p_button == action_button_action and is_instance_valid(picked_up_object):
		if picked_up_object.has_method("action_release"):
			picked_up_object.action_release()
		if picked_up_object.has_method("controller_action_release"):
			picked_up_object.controller_action_release(_controller)
```

Malcolmnixon (commented):
This is looking quite good - just a few minor checks and I think this will be ready to go in.

Malcolmnixon (approved):
This looks good to me

BastiaanOlij (approved):
Happy with the changes, great work.

PR #669, opened 2024-09-02
Added Poke docs link to interaction.md

tutek1:
In Features section -> Interaction Features, only the Pickup function and Pointer are mentioned. However, the toolkit supports the poke interaction, which is listed in the Interaction section.


Malcolmnixon (approved):
Looks good

PR #675, opened 2024-09-11
Update README.md regarding OpenXR plugin instructions

vigneshiamvk:
The current Readme says installation of OpenXR plugin is necessary for the demo to work, but that is only for older versions of Godot. Updated the instructions to reflect this.


BastiaanOlij (changes requested):
I think we can remove information about versions prior to Godot 4.0 as XR tools can't be used with older version of Godot.

We should however add some info about the need of the `godot_openxr_vendors` plugin if you want to export XR Tools games to Android based devices.

PR #679, opened 2024-09-20
rename grab state variable to avoid conflicts with tipical class_name 'State'

surreal6:
I suggest to rename this enum variable to avoid conflicts with anyone willing to use 'State' string as a class_name. 

This enum is used internally in this grab_driver and has no real impact to change it. But anyone trying to follow classic States Machine with nodes and defined classes, will find a conflict.


Malcolmnixon (approved):
Looks good and works well.

PR #683, opened 2024-09-27
Function gaze pointer

surreal6:
A new implementation of the Gaze Pointer but as an independent function. Based in #664 PR

With 'click on hold' boolean you can select to send clicks to the target node or just move the mouse around.

### DIFF
--- a/addons/godot-xr-tools/functions/function_gaze_pointer.gd
+++ b/addons/godot-xr-tools/functions/function_gaze_pointer.gd
@@ -1,0 +1,555 @@
+@tool
+@icon("res://addons/godot-xr-tools/editor/icons/function.svg")
+class_name XRToolsFunctionGazePointer
+extends Node3D
+
+
+## XR Tools Function Gaze Pointer Script
+##
+## This script implements a pointer function for a players camera. Pointer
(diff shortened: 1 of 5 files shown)


Malcolmnixon:
This is looking very good - although it did highlight a bug with the desktop integration overwriting the headset rotation even in full XR mode - https://github.com/GodotVR/godot-xr-tools/issues/684

Do you think it's beneficial to keep the laser functionality - it can be enabled and does work, but it looks so bad (the laser coming out between the eyes) I'm not sure it'd ever want to be enabled.

surreal6:
Oh, good, you found the bug with the desktop integration. I was unable to detect where it came from, but was suspecting from desktop integration and asked pietru about, but he didn't see any bug... 

Yes, i think we can remove the laser functionality.  As i understand, It its useful just to check if the pointer is working, but you can also use the target for this. I can remove the laser functionality for clearness.

surreal6:
this will fix #662 issue

RangerMauve:
TY for working on this!

Malcolmnixon:
With Pietru's fix this now works perfectly.

Malcolmnixon (approved):
This works very well. Nice job

PR #585, opened 2023-12-10
Added gizmos to help align interactable joints

Malcolmnixon:
This PR resolves request #584 by adding gizmos to align interactable hinges, sliders, and joysticks. This does cause minor breakage in that:
- New origin nodes have been introduced where the ranges are visualized
- Testing showed Y-Axis inversion on the slider and joystick

# Slider Gizmo
image

# Hinge Gizmo
image

# Joystick Gizmo
image

PR #685, opened 2024-10-03
Add SnapPath

squidt:
Extends from SnapZone and allows the user to specify a Path3D to which PickableObjects will be snapped. Snapping is done by creating a temporary SnapZone in-script along the closest point of the Path3D node. This temporary SnapZone is freed once the PickableObject is taken from the rail. Make snapping discrete with the "snap_interval", the distance between snaps in Meters.

image

### DIFF
--- a/addons/godot-xr-tools/objects/snap_path.gd
+++ b/addons/godot-xr-tools/objects/snap_path.gd
@@ -1,0 +1,105 @@
+@tool
+extends XRToolsSnapZone
+class_name XRToolsSnapPath
+
+## An [XRToolsSnapZone] that allows [XRToolsPickable] to be placed along a child [Path3D] node.
+##     They can either be placed along any point in the curve, or at discrete intervals
+##     by setting "snap_interval" above 0.0. Note: Attached [XRToolsPickable]s will face the +Z axis
+##
+## TODO:
+##     Feature: Preview Material that shows the held object on the rail
+##     Option: Blocking overlapping [XRToolsPickable] collisions
+##     Option: Discrete intervals by subdivision, not real space (Ex: 4 points equally spaced)
+
+
+## Real world distance between intervals in Meters.
+## Enabled when not 0
+@export  var snap_interval := 0.0:
+	set(v): snap_interval = absf(v)
+@onready var path : Path3D
+
+
+func _ready() -> void:
+	super._ready()
+	
+	for c in get_children():
+		if c is Path3D:
+			path = c
+			break
+
+
+func _get_configuration_warnings() -> PackedStringArray:
+	# Check for Path3D child
+	for c in get_children():
+		if c is Path3D:
+			path = c
+			return[]
+	return["This node has no path to place objects along. Consider adding a Path3D as a child to define its shape."]
+
+
+# Called when a target in our grab area is dropped
+func _on_target_dropped(target: Node3D) -> void:
+	# Skip if invalid
+	if !enabled or is_instance_valid(picked_up_object) or !is_instance_valid(target) or !path:
+		return
+	
+	# Make a zone that will destruct once its object has left
+	var zone = _make_temp_zone()
+	
+	# Set zone's transform to respect the rail
+	# If snap_interval has been set, use it
+	if snap_interval != 0.0:
+		var ideal_offset = _find_offset(path, target.global_position)
+		var s = snappedf(ideal_offset, snap_interval)
+		zone.transform = path.curve.sample_baked_with_rotation(s)
+	else:
+		zone.transform = path.curve.sample_baked_with_rotation(clamp(_find_offset(path, target.global_position), 0.0, 1.0))
+	
+	# Add zone as a child
+	path.add_child(zone)
+	zone.owner = path
+	
+	# Connect self-destruct with lambda
+	zone.has_dropped.connect(func(): zone.queue_free(), Object.ConnectFlags.CONNECT_ONE_SHOT)
+	
+	# Force pickup
+	if target.can_pick_up(self):
+		zone.pick_up_object(target)
+	else:
+		zone.queue_free()
+
+
+# Make a zone that dies on dropping objects
+func _make_temp_zone():
+	var zone = preload("res://addons/godot-xr-tools/objects/snap_zone.tscn").instantiate()
+	
+	# XRToolsSnapZone manaul copy
+	zone.enabled        = true
+	zone.stash_sound    = stash_sound
+	zone.grab_distance  = grab_distance
+	zone.snap_mode      = snap_mode
+	zone.snap_require   = snap_require
+	zone.snap_exclude   = snap_exclude
+	zone.grab_require   = grab_require
+	zone.grab_exclude   = grab_exclude
+	zone.initial_object = NodePath()
+	
+	# CollisionObject3D manual copy
+	zone.disable_mode       = disable_mode
+	zone.collision_layer    = collision_layer
+	zone.collision_mask     = collision_mask
+	zone.collision_priority = collision_priority
+	
+	return zone
+
+
+func _find_closest_point(_path: Path3D, _global_position: Vector3) -> Vector3:
+	# Transform target pos to local space
+	var local_position := global_position * _path.global_transform
+	return _path.curve.get_closest_point(local_position)
+
+
+func _find_offset(_path: Path3D, _global_position: Vector3) -> float:
+	# Transform target pos to local space
+	var local_pos: Vector3 = _global_position * _path.global_transform
+	return _path.curve.get_closest_offset(local_pos)
--- a/addons/godot-xr-tools/objects/snap_path.tscn
+++ b/addons/godot-xr-tools/objects/snap_path.tscn
@@ -1,0 +1,22 @@
+[gd_scene load_steps=4 format=3 uid="uid://dsstvanwd58r0"]
+
+[ext_resource type="Script" path="res://addons/godot-xr-tools/objects/snap_path.gd" id="1_m211o"]
+
+[sub_resource type="BoxShape3D" id="BoxShape3D_pik8g"]
+size = Vector3(0.1, 0.1, 1)
+
+[sub_resource type="Curve3D" id="Curve3D_w68am"]
+_data = {
+"points": PackedVector3Array(0, 0, 0, 0, 0, 0, 0, 0, -0.5, 0, 0, 0, 0, 0, 0, 0, 0, 0.5),
+"tilts": PackedFloat32Array(0, 0)
+}
+point_count = 2
+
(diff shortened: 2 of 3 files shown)


DigitalN8m4r3:
hmm, this looks really interesting...
most def gotta check this out, i wonder at the same time if this could be adapted for more usecases such
as slideable grabpoints which go along the line of the path and slideable magazines for firearms etc.
aynways, needs testing and as i can see there is no actual demo that comes with it?
you may want to add that so that the idea and functionality can be tested

DigitalN8m4r3:
> Extends from SnapZone and allows the user to specify a Path3D to which PickableObjects will be snapped. Snapping is done by creating a temporary SnapZone in-script along the closest point of the Path3D node. This temporary SnapZone is freed once the PickableObject is taken from the rail. Make snapping discrete with the "snap_interval", the distance between snaps in Meters.
> 

sorry but i just fail to get this working, could it be that something is missing or am i just not figuring this out the way its supposed to be...

squidt:
> sorry but i just fail to get this working, could it be that something is missing or am i just not figuring this out the way its supposed to be...

I'll work on adding a demo right away

DigitalN8m4r3:
just tested the updated version, lookin good for me
@Malcolmnixon you might want to give it a spin, it is a nice addition
@squidt not sure if you are interested but maybe you can follow up on my initial message regarding the expansion of this, anyways good work there!

squidt:
I accidentally made all these commits using 4.3.stable and I just noticed the repo is using 4.2 in the .project so they might not work for that version. "Parse Error: Saved with newer project format version". I'll try to fix that.
@DigitalN8m4r3 I have been interested in making an InteractablePath for objects like sniper bolts. I might work on that or something like a SnapArea that Pickables can 'float' in like many VR games have for backpacks.

squidt:
XRToolsPathGuide uses a Marker3D to define a length for XRToolsSnapPath to use so it can place the object fully on the path with no overhanging. Example added to the PicatinnyScope. The next feature I plan on adding would be using a pickable's existing collision shape as the shape used to pick it up off the rail instead of a circle that might be offset or wrongly sized. I believe there is a way to do that in-code, I'll try it later.

DigitalN8m4r3:
> XRToolsPathGuide uses a Marker3D to define a length for XRToolsSnapPath to use so it can place the object fully on the path with no overhanging. Example added to the PicatinnyScope. The next feature I plan on adding would be using a pickable's existing collision shape as the shape used to pick it up off the rail instead of a circle that might be offset or wrongly sized. I believe there is a way to do that in-code, I'll try it later.

any Chance of expqndong this to a optional enforcing to use the furthest point along a predefined axis/ selectable from x,y,z and if not reaching the closest point(closest to the pickable) = drop instead of snap. This could be useful for snapable mags.

squidt:
> any Chance of expqndong this to a optional enforcing to use the furthest point along a predefined axis/ selectable from x,y,z and if not reaching the closest point(closest to the pickable) = drop instead of snap. This could be useful for snapable mags.

Sorry I don't understand the question. I would like to make a solution for all the different types of magazines but that would be more appropriate to the 'interactable' classes where the user is driving a motion, then upon reaching the end the magazine is inserted. Examples such as an AK magazine where you have to rock it in, or a bolt handle you have to pull up to unlock before pulling back. I'll try to come up with a solution for that but it'll likely end up in a different pull request using a new class, where the final example may combine classes in the node tree to get a desired effect.

DigitalN8m4r3:
> > any Chance of expqndong this to a optional enforcing to use the furthest point along a predefined axis/ selectable from x,y,z and if not reaching the closest point(closest to the pickable) = drop instead of snap. This could be useful for snapable mags.
> 
> Sorry I don't understand the question. I would like to make a solution for all the different types of magazines but that would be more appropriate to the 'interactable' classes where the user is driving a motion, then upon reaching the end the magazine is inserted. Examples such as an AK magazine where you have to rock it in, or a bolt handle you have to pull up to unlock before pulling back. I'll try to come up with a solution for that but it'll likely end up in a different pull request using a new class, where the final example may combine classes in the node tree to get a desired effect.

Aah this is true, sorry that i bothered you but indeed your theoretical approach would be way better.
Anyways thanks for sharing your work and am lookin forward to testing it 😀

squidt:
> Aah this is true, sorry that i bothered you but indeed your theoretical approach would be way better. Anyways thanks for sharing your work and am lookin forward to testing it 😀

No worries you didn't bother me at all! It's gonna be fun seeing what can be done.

DigitalN8m4r3 (approved):
lgtm, tested in and out and it is not interfeering with anything.
even if my approval doesnt count when it comes to merging, i can at least say that i tested it and there are no issues!

addons/godot-xr-tools/objects/snap_path.gd
```diff
@@ -0,0 +1,155 @@
...
+	
+	# Add zone as a child
+	path.add_child(zone)
+	zone.owner = path
+	
+	# Connect self-destruct with lambda
+	zone.has_dropped.connect(func(): zone.queue_free(), Object.ConnectFlags.CONNECT_ONE_SHOT)
+	
+	# Force pickup
+	if target.can_pick_up(self):
```
BastiaanOlij, line 73: Little nitpick but shouldn't you check this at the start and not bother with creating the temporary zone if we can't pickup the object?

BastiaanOlij (approved):
Finally had time to have a closer look at the code side of this. Looks very clean!
Malcolm needs to have a quick pass over this but not sure if he'll have time quickly, but imho this can be merged as soon as he gives his ok.

addons/godot-xr-tools/objects/snap_path.gd
squidt: Yes, I missed that. I must've left it that way from when I first got it to work. Fixed in [472f0ae](https://github.com/GodotVR/godot-xr-tools/pull/685/commits/472f0aeea0441a519cef9fa28d4e5ad06096c22b)

squidt:
I just found out about gdlint and installed it. I'll make a commit to fix whatever it brings up.

squidt:
> any Chance of expqndong this to a optional enforcing to use the furthest point along a predefined axis/ selectable from x,y,z and if not reaching the closest point(closest to the pickable) = drop instead of snap. This could be useful for snapable mags.

I think I know what you mean by this now. Something like: A magazine path where the user brings a magazine close and it attaches to the path while close enough and still grabbed. The user then has to insert it fully to snap it or else 'when dropped' it will slide out/ fall. It sounds possible with the current classes I might work on making an example.

I'm experimenting more with the 'interactable' classes. I was able to make a charging handle that can be slid back until the end then tilted up to be locked. I could make a PR request implementing that for the sniper rifle but I did it with a custom slider class and a couple changes to the source of those classes.

It'll just depend on what I work on next, I've been prototyping a game with this plugin and I'll be trying to bring back any improvements or examples I find useful.

DigitalN8m4r3:
> > any Chance of expqndong this to a optional enforcing to use the furthest point along a predefined axis/ selectable from x,y,z and if not reaching the closest point(closest to the pickable) = drop instead of snap. This could be useful for snapable mags.
> 
> I think I know what you mean by this now. Something like: A magazine path where the user brings a magazine close and it attaches to the path while close enough and still grabbed. The user then has to insert it fully to snap it or else 'when dropped' it will slide out/ fall. It sounds possible with the current classes I might work on making an example.
> 
> I'm experimenting more with the 'interactable' classes. I was able to make a charging handle that can be slid back until the end then tilted up to be locked. I could make a PR request implementing that for the sniper rifle but I did it with a custom slider class and a couple changes to the source of those classes.
> 
> It'll just depend on what I work on next, I've been prototyping a game with this plugin and I'll be trying to bring back any improvements or examples I find useful.

Am sure this would be welcome, especialy since we are approaching the end of xr tools 1, v2 most probably wont include much thirdparty game specific stuff (i might be wrong but thats what i imagined from reading about it)

So yes, please go ahead if your solution works better then mine, am happy with mine getting ditched in favor of something more robust

squidt:
> So yes, please go ahead if your solution works better then mine, am happy with mine getting ditched in favor of something more robust

Thanks! it will be a breaking change but hopefully it ends up working better. The idea behind the custom slider is that the user can rotate the slider node with the Editor's rotation gizmos and using a 'private Transform3D' we save an origin point and rotation whenever there is a transform change from non-interactable-driven-behavior. Then the slider always slides 'back' from however its been rotated. So far I have it working for the slider and hinge.

The benefits are that there's no more SliderOrigin HingeOrigin nodes required, and the setup feels more intuitive from a user perspective, rotating the slider node into place rather than changing the surrounding nodes to line up with slider behavior. I'll make a branch later this week and a PR when it's got all the parts together.

DigitalN8m4r3:
> > So yes, please go ahead if your solution works better then mine, am happy with mine getting ditched in favor of something more robust
> 
> Thanks! it will be a breaking change but hopefully it ends up working better. The idea behind the custom slider is that the user can rotate the slider node with the Editor's rotation gizmos and using a 'private Transform3D' we save an origin point and rotation whenever there is a transform change from non-interactable-driven-behavior. Then the slider always slides 'back' from however its been rotated. So far I have it working for the slider and hinge.
> 
> The benefits are that there's no more SliderOrigin HingeOrigin nodes required, and the setup feels more intuitive from a user perspective, rotating the slider node into place rather than changing the surrounding nodes to line up with slider behavior. I'll make a branch later this week and a PR when it's got all the parts together.

cant wait to check it out, its possible that it will need readjustment after the following pr get merged #585

DigitalN8m4r3:
@squidt 
this one has like nothing to do with the above but would be a interesting thing.

So bare with me as this is all just smoke and mirror theory about Stabbing,
as in moving an object forcefully towards a body. Upon impact
, taking the velocity into account and the bodys physical attributes as in is it heavy/medium/soft to 
determine how much force/Power = velocity is needed to penetrate the body.

Am thinking about using a path based solution to snap the pickable upon penetration inside the body while still being able to hold the pickable but using the physical attributes to determine how easy or hard it is to move the pickable out/further out/in the body.
Now the last part might be earliest possible once we get weight feel implemented into xr tools but the rest might verry well be already doable.

Not sure when but i would love to try my hands at it just thought i would ask you about your opinion and maybe its something you already have thought about.
Btw you should be on the discord, would make it easier to annoy you! 😉

squidt:
@DigitalN8m4r3 Yea I could see that, we just need some way to keep a grab active while attaching to a path or an interactable then you can do so much including that. I'm in the mux213 discord in general chat as 'squid' if you wanna bother feel free, or the official godot discord.

BastiaanOlij:
Malcolm unfortunately is currently swamped with work, I don't want to leave this as this is good work. We can always improve on it later. So I am merging this.

PR #686, opened 2024-10-03
Fix 'create_snap' desync

squidt:
If the player drops then picks up a 'SnapZone' with a 'PickableObject' in it the 'PickableObject' will lag behind during Player or SnapZone movement due to calling 'create_snap'. Adding the driver as a direct child of the grabber prevents any desync. 'create_lerp' function is similar but has been left out until further testing.

Tested on Godot 4.3.stable

### DIFF
--- a/addons/godot-xr-tools/objects/grab_points/grab_driver.gd
+++ b/addons/godot-xr-tools/objects/grab_points/grab_driver.gd
@@ -199,9 +199,8 @@ static func create_snap(
 	# Snapped to grab-point so report arrived
 	p_grab.set_arrived()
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):
Would like to get some feedback from others to make sure this doesn't have any unexpected side-effects. Other than that, I think this is safe to merge.

DigitalN8m4r3 (approved):
finally got around, lgtm

BastiaanOlij:
This needs a rebase and gdlint errors fixed, and then I will merge it :)

squidt:
Think I rebased right.

Issue #684, opened 2024-09-29
Desktop Turn overwriting camera X-Rotation even in full XR mode


Malcolmnixon:
The movement_desktop_turn.gd scripts physics_movement() method is resetting the player body camera X-rotation to 0 even when XR is actively being driven by a headset.
https://github.com/GodotVR/godot-xr-tools/blob/7c6d455b7d6f4b05e4874cab4d53912270a3d84b/addons/godot-xr-tools/desktop-support/movement_desktop_turn.gd#L87-L95

When not in desktop mode the desktop movement providers should do nothing.

Malcolmnixon:
@pietru2004 Could you look into this?

pietru2004:
Last time I checkedit wasnn't doing anything when xr was active(headset active). Gona check again...

pietru2004:
If you set ` clear_cam_x_when_body_not_active` to false there ahould be no problems, also I think problem that surreal6 has is related to vr headset he had connected to pc.

Malcolmnixon:
Yes, real XR mode can be fixed by overriding the default 'true' value of that value to be false, but that means the user has to explicitly disable a portion of the desktop behavior to get XR mode functioning correctly.

I'm still not clear why the desktop code is doing anything when the user isn't in desktop mode.

surreal6:
Hi there, this seems to be the bug I was experimenting with the desktop function and the gaze pointer. I can confirm the problem

pietru2004:
hm... If I recall correctly the `clear_cam_x_when_body_not_active` was made for sitting and was meant only for desktop so the if statement should probably be altered. The clear_mouse_move_when_body_not_active is supposed to clear any mouse movement if player is not in desktop mode

PR #687, opened 2024-10-03
Update movement_desktop_turn.gd

pietru2004:
This might fix https://github.com/GodotVR/godot-xr-tools/issues/684
or atleast it is an atempt - bit higher in `func _process(_delta: float) -> void:` there is similar code that hopefully should work as intended...

Note: My VR headset is atm not functional so I can't test it... Graphic drivers issue.


pietru2004:
If I remember correctly
```
	if clear_cam_x_when_body_not_active:
		player_body.camera_node.rotation_degrees.x=0
```
was meant for [sitting](https://github.com/GodotVR/godot-xr-tools/pull/462) so I think it should be safe to just either comment out it in physics_movement or set it to false by default

Malcolmnixon (approved):
This looks good to me and appears to work fine.

PR #692, opened 2024-11-17
Rework interactables

squidt:
Interactables use an empty Node3D as a parent and 'origin' node to set up both handles and interactable_handle_driven scenes. This was unintuitive for new users (me). This PR will attempt to make the scenes stand alone and more intuitive. Sliders can be orientated in the editor and slide along their rotated axis. Hinges can do the same and offer 'hinge_axis' as a variable to allow the user to specify an axis to rotate around.

These are breaking changes and a work in progress. Joysticks must be incorporated and handles tested. I'll adjust the interactable_demo_scene to feature the final changes.

### DIFF
--- a/addons/godot-xr-tools/interactables/interactable_handle.gd
+++ b/addons/godot-xr-tools/interactables/interactable_handle.gd
@@ -23,13 +23,13 @@ extends XRToolsPickable
 @export var snap_distance : float = 0.3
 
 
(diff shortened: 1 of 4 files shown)


squidt:
Last commit also adding 'hinge_wrapping' bool for wrapping between min and max limits (infinitly spinning wheel)

PR #695, opened 2024-11-21
Add weighted logic to collision hands

BastiaanOlij:
This is an implementation of weighted hands based on the current logic for holding objects in XR Tools.
It's far from perfect but it does a decent job within the current constraints of XR tools.

The main missing bit that would be an improvement is missing angular momentum when an object is held in a single hand (as dual wielding already compensates to an extent). But in order to do so we really need to tie properly into the physics engine and that requires a completely different type of logic for holding objects (I have been doing experimentation for this and may implement that approach for XR tools 2).

The other shortcoming is that when the user teleports, the movement breaks the maximum distance and we let go of the object. This I plan to resolve in a follow up PR as this requires proper signals being hooked up to our teleport system and I'm already planning work on this.

*Contributed by Khronos Group through the Godot Integration Project*


DigitalN8m4r3 (commented):
setting max pickup force on the collision hands to 250 gives better weight feel, however it somewhat feels disconnected.
for a sniper rifle, an empty one without scope would weight around 5.4kg/12lb.
snap rotation is not working properly with weight, the object lags behind and makes the hand that got the object picked up, lag behind in a twitching way.
not sure if its related but i cant properly grab the sniper handle, it disconnects as in not being able to fully grab it.
will have to play more with this pr but it is a good start.

BastiaanOlij:
> not sure if its related but i cant properly grab the sniper handle, it disconnects as in not being able to fully grab it.
> will have to play more with this pr but it is a good start.

If the object you pick up is too heavy that it lags behind too much, we'll eventually let it go. One thing I do need to fix here is that player movement is added into this as well.

BastiaanOlij:
Player movement is now directly applied to collision hands (and poke body while I was in there) including when you use teleporting.

The only scenario where we're not automatically reacting is if the player is moved outside of the player body methods (a few movement providers do this) but I need to react differently depending on how the player is moved.

DigitalN8m4r3:
> Player movement is now directly applied to collision hands (and poke body while I was in there) including when you use teleporting.
> 
> The only scenario where we're not automatically reacting is if the player is moved outside of the player body methods (a few movement providers do this) but I need to react differently depending on how the player is moved.

right now even with your settings its not acceptable, upon performing turn movement it twitches and it seems that your recent changes made the still movement of the hands when holding, twitch just a lil as well and i think this wasnt the case before.
Too bad this really needs another pair of eyes, one that can actualy suggest code changes instead of my 8ball comments

BastiaanOlij:
Hmm, I wonder why its that much more stable for me and what I've missed.
Turning and moving shouldn't have an effect anymore because we no longer apply the weight simulation. Same with teleport.
So there must be some scenarios where we're getting some oscillation.

Malcolm also noticed that because we're now properly testing for collisions, we're getting into trouble when we attach items to snap zones that bring collisions close to the hands, such as adding the scope to the gun.

DigitalN8m4r3:
> Hmm, I wonder why its that much more stable for me and what I've missed. Turning and moving shouldn't have an effect anymore because we no longer apply the weight simulation. Same with teleport. So there must be some scenarios where we're getting some oscillation.
> 
> Malcolm also noticed that because we're now properly testing for collisions, we're getting into trouble when we attach items to snap zones that bring collisions close to the hands, such as adding the scope to the gun.

indeed, its a hornets Nest.. 
what i have noticed is that indeed the interactables on the sniper are malfunctioning most of the time, just not when the sniper is hoizontal from the side.

BastiaanOlij:
OK, as far as I can tell the issues we were having had to do with self colliding between our collision hands logic and the objects we had picked up, especially if those object had attachments.

As anything that is picked up or attached is added to physics layer 17, and there is no point for the  hands to collide with what we've picked up, I've removed layer 17 from the layer mask on collision hands.

That seems to have solved some of the weird issues.

I've also added an extra set of hands to visualise where our hands actually are located. This needs a little more tweaking as it doesn't work right when we're using specific grip points for the left or right hands as we hide things but thats something to tweak a little more later.

DigitalN8m4r3 (approved):
from my pov i see this as complete, maybe a nudge here and there but all in all its probably the most stable we can get
and here is a test of the pr using turn - smooth compared to turn snap, you will see the remaining issue and while i can live without a fix for snapturn, some might wish for a fix, however i would like to express here that this solution is simply not going to make evryone happy, since at the same time other users might wish for collision on held objects towards the hands when holding but this scenario is a hornets nest, anyways here is my test

https://github.com/user-attachments/assets/d1583cba-0a6d-46c8-9639-27398af656ff

BastiaanOlij:
I've set this ready for review because I think this is as good as I'll get it for now.
I'm still at a loss to explain the issue with snap rotation and the held object seeming to lag. I think it may have something to do with the remote transform updating the remote nodes position incorrectly.
Seeing we already handle all the logic here I wonder if it even makes sense to use RemoteTransform here and not just apply that logic ourselves.

addons/godot-xr-tools/objects/grab_points/grab_driver.gd
```diff
@@ -30,6 +30,8 @@ var lerp_duration : float = 1.0
 ## Lerp time
 var lerp_time : float = 0.0
 
+## Player body
```
BastiaanOlij, line 33: Note to self, either `RemoteTransform3D` is not updating in time, or this is not our fix and doubling up on the move signals from our Player Body is overkill. Might remove changes to this file.

DigitalN8m4r3 (approved):
as mentioned on discord, this looks good now and the extra measures taken are well Done.
We know this can not get much better, givin ppl a simple note that this doesnt work with snap turn should be enough.
Lookin foward to @Malcolmnixon Review

addons/godot-xr-tools/objects/grab_points/grab_driver.gd
BastiaanOlij: Removed the code that is overkill.

Malcolmnixon (approved):

Issue #613, opened 2024-01-08
Issue with pickable objects when scaling is different


elisagalofaro:
Hi there! I'd like to bring to your attention an issue I came across in my project. 
Specifically, when I have a Node3D (referred to as Test01 in the illustration) with a scale of, let's say, 5, and I instantiate a pickable object (depicted as PickableObject in the figure) as its child with a scale of 1, the grabbed item tends to move away from the hand, exhibiting an offset proportional to the scale. 
Unfortunately, I'm unable to adjust the scaling in my tree due to the complexity of the model in my project.
Thank you in advance for the help!

![test01](https://github.com/GodotVR/godot-xr-tools/assets/56263531/0f5adbf7-488b-412a-8b1b-8b8ebef5fcd3)
![pickableObject](https://github.com/GodotVR/godot-xr-tools/assets/56263531/94c812b6-98ca-49c9-91ce-7f4f28925808)

davehakim:
I see the same. I notice that if I modify `grab_driver.gd` and `grab.gd` to replace all the calls to `transform.inverse` with `transform.affine_inverse` the pickable nodes stop jumping away from the hands. I forget but adding `driver.update_scale = false` may also be necessary in the `create_snap` method.

However once the parent of the pickable is both scaled and rotated even the patch above broke down for me.

davehakim:
Actually, I just retried updating all the calls to `transform.inverse` in `grab_driver.gd` and `grab.gd` to  `transform.affine_inverse` locally and re-ran my project and that alone seems to fix the issues I had with XRToolsPickables nested under a scaled + rotated nodes. I could try to spin up a pr for this @BastiaanOlij if the pickable object code hasn't migrated too far away from what I'm using in godot 4.3

PR #707, opened 2025-02-08
Fix grab with scaled pickables by swapping inverse for affine_inverse

davehakim:
[Jolt physics](https://github.com/godot-jolt/godot-jolt) appears to support the scaling of physics bodies in response to the scaling of an ancestor node. This has great potential for tabletop XR games where one might want the pieces to be 'pickable' (a subclass of `RigidBody3D`) while at the same time allowing dynamic scaling of the game board (and all the pieces on it!)

The grab related classes however don't allow grabbing scaled `XRToolsPickable`s by default (well, at very least the grab position is off) because they use `inverse` transforms and not `affine_inverse` which respects node scaling. This PR changes all the relevant `inverse` transforms to `affine_inverse` which should fix that.

To test:

1. Using godot 4.3 add the Godot Jolt framework 
2. Choose jolt as the 3d physics engine for the demo project
3. Run the project and enter the pickables scene

Without the changes in this pr the cubes would not be pickable (they disappear or shrink in response to attempted pickups)
With the change they pick up fine (usually 🤷) at the correct grab position

Issues:

- This pr mostly works but not 100% of the time 😂 . For reasons I have yet to discover (which is part of why I'm opening this draft PR! To get assistance from developers who know this code better than I do!) sometimes what I think should occur as a normal pickup seems to happen as a ranged pickup.
- This pr does not (yet) work in the case of ranged pickups.   

Note 1 : The changes to pickable_demo are only here to illustrate the issue and need to be cleaned up or out before this pr gets considered for merge.
Note 2 : I expect this addresses the cause of https://github.com/GodotVR/godot-xr-tools/issues/613

To work around the 'unexpected scaling' glitch in my game I've just disabled ranged grabs (by always treating ranged grabs in the code the same as non ranged i.e. always doing this in the pickup function

```
var grab := Grab.new(grabber, self, by_grab_point, true)
```

but I figure before this pr becomes something more general purpose though the ranged grab issue(s) needs to be solved

### DIFF
--- a/addons/godot-xr-tools/objects/grab_points/grab.gd
+++ b/addons/godot-xr-tools/objects/grab_points/grab.gd
@@ -66,7 +66,7 @@ func _init(
 	if p_point:
 		transform = p_point.transform
 	elif p_precise:
-		transform = p_what.global_transform.inverse() * by.global_transform
+		transform = p_what.global_transform.affine_inverse() * by.global_transform
 	else:
 		transform = Transform3D.IDENTITY
 
--- a/addons/godot-xr-tools/objects/grab_points/grab_driver.gd
+++ b/addons/godot-xr-tools/objects/grab_points/grab_driver.gd
@@ -37,7 +37,7 @@ func _physics_process(delta : float) -> void:
 		return
 
 	# Set destination from primary grab
-	var destination := primary.by.global_transform * primary.transform.inverse()
+	var destination := primary.by.global_transform * primary.transform.affine_inverse()
 
 	# If present, apply secondary-node contributions
 	if is_instance_valid(secondary):
@@ -47,7 +47,7 @@ func _physics_process(delta : float) -> void:
 
 		# Calculate the transform from secondary grab
 		var x1 := destination
-		var x2 := secondary.by.global_transform * secondary.transform.inverse()
+		var x2 := secondary.by.global_transform * secondary.transform.affine_inverse()
 
 		# Independently lerp the angle and position
 		destination = Transform3D(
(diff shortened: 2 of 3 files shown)


BastiaanOlij (approved):
This is good, we should merge this. If there is a chance to revert the changes on pickable_demo.tscn as they are not part of the fix, that would be great.

davehakim:
> This is good, we should merge this. If there is a chance to revert the changes on pickable_demo.tscn as they are not part of the fix, that would be great.

Hi @BastiaanOlij , sorry for the delay in circling back to this. I've reverted the demo changes so the PR _could_ merge as is and it will fix scaling issues when directly grabbing an XRToolsPickable . Two things tho

1. LERP ranged grabs still need fixes that I haven't looked into yet with this pr (because my game doesn't use them ... and I don't have a great understanding of LERPing 😁 )
2. Snap ranged grabs appear to need this change which isn't yet in my PR (since I don't understand the full impact of it)

```
var grab := Grab.new(grabber, self, by_grab_point, true)
``` 

2a. I see some non-ranged grabs executing as ranged grabs (it looks like `picked_up_ranged` on `XRToolsFunctionPickup` can be true even in some cases when `can_ranged_grab` on the XRToolsPickable is not) so even in some cases of non-ranged grabs we can get some scaling issues unless I make the change listed in 2

So what do you think @BastiaanOlij 

a. Get this pr in as is ?
b. Update ranged grab snap case to set `precise` Grab argument to true and then look to merge ?
c. Or don't attempt to merge until we solve ranged LERP grabs (which I'd need help to do 😁 )

BastiaanOlij:
Thanks @davehakim , I'm happy to merge this as is and see further improvements in coming PRs.

PR #709, opened 2025-03-06
Upgrade to 4.4 and run Upgrade uids

jamie-pate:
4.4 is viral and any addons also need uids checked in :(


BastiaanOlij:
We are not doing this any time soon, there are still plenty of people on 4.3.

We will do this eventually

jamie-pate:
The pain is real with 4.4 though, maybe you could add the .uid files without updating the version number?

pietru2004:
Please don't update to 4.4 UIDs system... they are pain
I stay on 4.3 due to 4.4 UID files...

Also there is whole thread of people disliking them and there no being alternative options...

BastiaanOlij:
@jamie-pate totally agree, it is a major PITA for us plugin developers.

It gets even worse, with XRTools 2 that I'm working on, I'm copying code across from XR tools 1 and thus copying UIDs across. Causes nightmares when you include both plugins in a transitional period as now you have duplicates Godot "fixes" by messing up your entire project.

I solved it in XRT2 by adding a CI script that forces all UIDs to start with `xrt2`, it's a PITA but atleast it ensures that its possible to include both plugins in a project.

I'm tempted to do the same with XR Tools 1 once we decide to make the minimum version Godot 4.4 and indeed introduce UIDs.
The pain of changing UIDs is unfortunately unavoidable but I want to be really careful on when we introduce it.

BastiaanOlij:
Now that Godot 4.5 is out, we had a discussion in the last XR meeting to bring the minimum version of XR Tools up to 4.4.

If you have time to rebase this that would be great. Else I'm happy to redo this from my end.

PR #710, opened 2025-03-10
Improve performance by avoiding force_update_transform on grab driver

DanielKinsman:
I have a scene with hundreds of pickables tucked away in snap zones waiting for the player do do something with them. The repeated calls to `force_update_transform()` every physics step keeps them from sleeping and the performance tanks.

This change improves it from <10fps to 90fps on the XR2. A quick test in the pickables demo shows no changes to behaviour.


BastiaanOlij (approved):
Approved in the XR meeting.

PR #714, opened 2025-03-21
Improved material logic in Viewport2Din3D

### LINKED ISSUE #708: Improve material handling of Viewport2Din3D
A while ago we changed the material handling in Viewport2Din3D to allow users to supply a material.

Unfortunately due to the need to assign the ViewportTexture we needed to add code that duplicated that material, and likely due to us not being able to subscribe to changes on that material, we duplicate the material at intervals when we're in the editor.

If no material is provided, we create our own but this same enhancement seems to have resulted in that material being recreated over and over while the editor is used.

This logic needs to be improved, instead of re-creating a material we can update the material we've created with the new settings. If no settings have changed this should result in nothing being done.

While we're at it, we should also investigate if we can use a static repository of materials if multiple Viewport2Din3D are in play that all use the same material settings (as only the Viewport Texture will be different). This should greatly cut down overhead when Viewport2Din3D is used a lot.

BastiaanOlij:
We have logic in our Viewport2Din3D to create the material regularly in the editor because we don't have a way to detect material changes. This logic is also the way it is because we can't use the source material directly as we are modifying it with the ViewportTexture being used.

The logic was also resulting in the material being recreated if no custom material was specified.

This PR changes the logic so that:
- We only duplicate our custom material when it is set
- We only copy a custom material once or after our custom material is cleared
- We will copy any change properties from our custom material to our screen material when required

This should ensure that in most use cases, we only create our shader once and reuse shaders as much as we can.

Fixes #708

### DIFF
--- a/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
+++ b/addons/godot-xr-tools/objects/viewport_2d_in_3d.gd
@@ -97,7 +97,7 @@ const DEFAULT_LAYER := 0b0000_0000_0101_0000_0000_0000_0000_0001
 @export var material : StandardMaterial3D = null: set = set_material
 
 ## Transparent property
-@export var transparent : TransparancyMode = TransparancyMode.TRANSPARENT: set = set_transparent
+var transparent : TransparancyMode = TransparancyMode.TRANSPARENT: set = set_transparent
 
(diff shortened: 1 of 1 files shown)


Malcolmnixon (approved):
This looks good.

PR #716, opened 2025-04-03
Gracefully handle freed objects hanging around as collision exceptions

DanielKinsman:
Freeing bodies that exist in other bodies' collision exceptions is a bit of a nightmare, see https://github.com/godotengine/godot/pull/104904

This works around it, though you will still see a few errors in the output during the 0.5 second timeout before they are cleaned up. We could try and keep track of the exceptions we've added on both sides of the fence and clean them up when freeing them, but that could get messy.

In my actual game I will probably avoid freeing them until the exceptions are cleared but this will allow me (and others) diagnose the problem and handle it more gracefully when it occurs, rather than having the output spammed with infinite error messages.


BastiaanOlij:
At first glance this looks solid. Needs proper testing.

addons/godot-xr-tools/objects/grab_points/grab.gd
```diff
@@ -229,7 +226,6 @@ static func _remove_collision_exceptions( \
 
 	# For now we'll remove all.
 
-	for body : PhysicsBody3D in exceptions:
-		if is_instance_valid(body):
-			on_collision_hand.remove_collision_exception_with(body)
-			body.remove_collision_exception_with(on_collision_hand)
+	for body : RID in exceptions:
```
BastiaanOlij, line 229: If we are going to remove the collisions directly on the physics server, we should also add them this way.

PR #724, opened 2025-05-02
fix some value is null cases

pietru2004:
I was testing my game and got variable is null errors at
`var obj = with.get_collider()`
`ground_node = ground_collision.get_collider()`
`var collision_node := collision.get_collider()`

Added `variable!=null` checks before code parts that would use those...

### DIFF
--- a/addons/godot-xr-tools/player/player_body.gd
+++ b/addons/godot-xr-tools/player/player_body.gd
@@ -421,7 +421,7 @@ func move_player(p_velocity: Vector3) -> Vector3:
 			var with = get_slide_collision(idx)
 			var obj = with.get_collider()
 
-			if obj.is_class("RigidBody3D"):
+			if obj!=null and obj.is_class("RigidBody3D"):
 				var rb : RigidBody3D = obj
 
 				# Get our relative impact velocity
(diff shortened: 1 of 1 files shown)


addons/godot-xr-tools/player/player_body.gd
```diff
@@ -421,7 +421,7 @@ func move_player(p_velocity: Vector3) -> Vector3:
 			var with = get_slide_collision(idx)
 			var obj = with.get_collider()
 
-			if obj.is_class("RigidBody3D"):
+			if obj!=null and obj.is_class("RigidBody3D"):
```
BastiaanOlij, line 424: Just change this to `if obj and obj.is_class("RigidBody3D")`, reads easier and is correct gdscript syntax.

addons/godot-xr-tools/player/player_body.gd
```diff
@@ -678,25 +678,26 @@ func _update_ground_information(delta: float):
 	ground_angle = rad_to_deg(ground_collision.get_angle(0, up_gravity))
 	ground_node = ground_collision.get_collider()
 
-	# Select the ground physics
-	var physics_node := ground_node.get_node_or_null("GroundPhysics") as XRToolsGroundPhysics
-	ground_physics = XRToolsGroundPhysics.get_physics(physics_node, default_physics)
+	if ground_node!=null:
```
BastiaanOlij, line 681: Same here, just change to `if ground_node:`

BastiaanOlij: (though if I am not mistaken, `XRToolsGroundPhysics.get_physics` already does a null check and returns default_physics if physics_node is not usable, so possibly this fix is incorrect.

addons/godot-xr-tools/player/player_body.gd
```diff
@@ -763,26 +764,27 @@ func _apply_velocity_and_control(delta: float):
...
-		var bounciness := XRToolsGroundPhysicsSettings.get_bounciness(
-				collision_physics, default_physics)
-		var magnitude := -collision.get_normal().dot(local_velocity)
-
-		# Detect if bounce should be performed
-		if bounciness > 0.0 and magnitude >= bounce_threshold:
-			local_velocity += 2 * collision.get_normal() * magnitude * bounciness
-			velocity = local_velocity + ground_velocity
-			emit_signal("player_bounced", collision_node, magnitude)
+		if collision_node!=null:
```
BastiaanOlij, line 767: Also `if collision_node:`

BastiaanOlij (changes requested):
Other than some nitpicks on the syntax, this looks  good. I do question the ground physics change though, might need @Malcolmnixon to weigh in.

addons/godot-xr-tools/player/player_body.gd
pietru2004: IDK about `XRToolsGroundPhysics.get_physics`, but when I was developing my project it was one of points where I got null exception so mayby it failed to do so ?

pietru2004:
removed those 3 `!=null`

Issue #670, opened 2024-09-05
XRToolsPointerEvent doesn't appear to be documented?


Meorge:
Searching for "XRToolsPointerEvent" on https://godotvr.github.io/godot-xr-tools/docs/ didn't bring up any results (in fact, the page didn't seem to load *any* search results... I might have to open another issue for that). I would have thought that XRToolsPointerEvent would have been documented in https://godotvr.github.io/godot-xr-tools/docs/pointer/, but I only see docs for XRToolsFunctionPointer, XRToolsInteractableBody, XRToolsInteractableArea, and Viewport2Din3D.

c-mahn:
I would also like to add myself to the list of this issue.

The Documentation states, that I would have the following Signals when using for example XRToolsViewport2DIn3D:
Signal 	Description
pointer_entered(body) 	The pointer entered this body/area
pointer_exited(body) 	The pointer no longer points at this body/area
pointer_moved(body, from, to) 	The pointer moved while pointing at this body/area*
pointer_pressed(body, at) 	The button was pressed while pointing at this body/area
pointer_released(body, at) 	The button was released*

But I only have "pointer_event(event: Variant)" from which I can't seem to grasp what to do. This is not documented. I have tried following the code to see, what I get, but I'm stuck with the following information:

- It's an Object because it's of type "24"
- The properties are as followed:
```
{ "name": "event_type", "class_name": &"XRToolsPointerEvent.Type", "type": 2, "hint": 0, "hint_string": "", "usage": 69632 }
{ "name": "pointer", "class_name": &"Node3D", "type": 24, "hint": 0, "hint_string": "", "usage": 4096 }
{ "name": "target", "class_name": &"Node3D", "type": 24, "hint": 0, "hint_string": "", "usage": 4096 }
{ "name": "position", "class_name": &"", "type": 9, "hint": 0, "hint_string": "", "usage": 4096 }
{ "name": "last_position", "class_name": &"", "type": 9, "hint": 0, "hint_string": "", "usage": 4096 }
{ "name": "RefCounted", "class_name": &"", "type": 0, "hint": 0, "hint_string": "RefCounted", "usage": 128 }
{ "name": "script", "class_name": &"Script", "type": 24, "hint": 17, "hint_string": "Script", "usage": 1048582 }
[...]

PR #733, opened 2025-06-14
Added new Signal to Function Pointer

c-mahn:
This commit adds a new signal to the Function Pointer, that reports back, if the FunctionPointer is interacting with something like a menu.


c-mahn:
I would like to add, that this is way simpler to understand than the undocumented XRToolsPointerEvent #670

PR #738, opened 2025-07-17
Fix #713: rename ControlerHider → ControllerHider

### LINKED ISSUE #713: `ControlerHider` is misspelled
`ControlerHider` should be `ControllerHider`.

(Minor cosmetic issue, not worth a major).

Suriya-Palaniswami:
Renamed ControlerHider to Controller Hider

Tested this by importing it to Godot 4.3. No output errors and the editor compiles successfully. 

XRToolsDesktopControllerHider is compiled as seen in the screenshot. 

godot#173

### DIFF
--- a/addons/godot-xr-tools/desktop-support/controler_hider.gd
+++ b/addons/godot-xr-tools/desktop-support/controler_hider.gd
@@ -1,11 +1,11 @@
 @tool
 @icon("res://addons/godot-xr-tools/editor/icons/function.svg")
-class_name XRToolsDesktopControlerHider
+class_name XRToolsDesktopControllerHider
 extends Node
 
-## XR Tools Controler Hider
+## XR Tools Controller Hider
 ##
-## This script hides controler if XR is not active.
+## This script hides controller if XR is not active.
 
 var _pointer_disabler := false
 var _last_xr_active := true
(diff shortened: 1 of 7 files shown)


BastiaanOlij:
Would need to test if this causes a backwards compatibility issue for project already using this, but I don't think so as the file name doesn't change.

Would be nice if some of the accidental changes could be filtered out (changes to the action map and positioning in the main level scene). But that's probably not that easy so more of a nitpick.

BastiaanOlij (approved):
I think this is ok but would like a second opinion from @pietru2004

pietru2004:
I don't see a problem, but please keep in mind that that it should be probably marked in **red** in changelog or smthg like that since class name changes here.

Also this one will probably only be important to remember for those that use desktop support.

pietru2004:
also is there difference between `controler` and `controller` ?

Suriya-Palaniswami:
@pietru2004 Sorry, I am new here. Not sure if I should be the one updating the changelog? And to answer your question, there is no difference between controler and controller. It was just misspelled.

pietru2004:
to be honest @Suriya-Palaniswami IDK should you be updating changelog, I assume the core maintainers will do it before release or smthg like that

BastiaanOlij:
We try and update the change log on release, but if its added to the PR than that just takes away the effort :)

BastiaanOlij:
@Suriya-Palaniswami if you don't mind giving this a quick rebase I'll merge it in.

BastiaanOlij:
I've rebased this into #757 and merged, thanks!

PR #740, opened 2025-08-03
Use correct up direction in player movement

tomaaron:
When the player walks up a slope, it gets with launched into the air with enough speed. This occurs because the `move_body` function sets the `up_direction` to the `ground_vector`, where this is the surface normal. That leads to incorrect physics calculation.

The proposed change is to use the gravity-based up direction (`up_gravity`) instead.

### DIFF
--- a/addons/godot-xr-tools/player/player_body.gd
+++ b/addons/godot-xr-tools/player/player_body.gd
@@ -399,7 +399,7 @@ func request_jump(skip_jump_velocity := false):
 func move_player(p_velocity: Vector3) -> Vector3:
 	velocity = p_velocity
 	max_slides = 4
-	up_direction = ground_vector
+	up_direction = up_gravity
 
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Make sure to test this with the [sphere world demo](https://github.com/GodotVR/godot-xr-tools/tree/master/scenes/sphere_world_demo), I have a feeling this fix will break that.

No idea why we suddenly get this launch behaviour. @Malcolmnixon any ideas?

tomaaron:
> Make sure to test this with the [sphere world demo](https://github.com/GodotVR/godot-xr-tools/tree/master/scenes/sphere_world_demo), I have a feeling this fix will break that.
> 
> No idea why we suddenly get this launch behaviour. @Malcolmnixon any ideas?

Thanks for getting back. I've tested it with the sphere world and it works the same.

PR #744, opened 2025-10-28
Support using Grip and Palm pose for hands

BastiaanOlij:
As more and more runtimes are using the aim pose as originally intended, e.g. the direction is no longer straight ahead but what comes naturally for UI and object interaction on a platform, we're moving to supporting grip and palm pose as well.

The demos are also being updated to use the grip pose, though I may do a few demos in a separate PR just to keep this PR easier to review.

Note that you'll find a new `hand_offset_mode` property on a number of nodes that relate to hand positioning (our hand meshes, pickup function, etc). The default setting here is "auto" and it will try and figure out what pose you're using based on what's set on the controller. This means that if you're still using the aim pose, things mostly stay the same (a lot of things will now be auto positioned on the palm which they should have been from day one). If however you switch to grip, everything should nicely adjust itself.

> [!NOTE]
> The `auto` logic assumes you are using the default action map. If you've created your own pose actions, the logic doesn't know what pose this is bound to. In this case you can select `aim`, `grip` or `palm` to tell the logic which type of pose you're expecting to be used.

> [!IMPORTANT]
> Because the grip pose has a 45 degree rotation, when you use the grip pose for your controller, all the effected children will be rotated in the opposite direction. This may look weird when setting up your scene. You will see that in all demo projects I've already updated, I've rotated the `XRController3D` nodes by 45 degrees as they would be in runtime.

### DIFF
--- a/.gdlintrc
+++ b/.gdlintrc
@@ -1,0 +1,1 @@
+max-line-length: 200
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,4 @@
 *.tres eol=lf
 *.import eol=lf
 *.gdshader eol=lf
+*.uid eol=lf
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -8,6 +8,7 @@
 - Fix grab with scaled pickables by swapping inverse for affine_inverse
 - Fix "UP" direction in player movement
 - Gracefully handle freed objects with collision exceptions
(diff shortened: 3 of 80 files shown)


BastiaanOlij:
Ok all, I'm going to merge this and see what the feedback will be. I'm planning on spending a few more days testing and improving the various demos before we work towards our long overdue 4.5.0 release.

PR #747, opened 2025-11-12
Fix gdlint errors, warnings.

squidt:
This PR is broken up into several commits that can be cherry-picked or excluded as needed.

- Remove semi-colons from .gd
- Replace semi-colons in comments with commas where appropriate
- Fix shadowed variables warnings
- Fix missing `@tool` in classes that extend from `@tool` scripts (MISSING_TOOL warning).

### DIFF
--- a/addons/godot-xr-tools/user_settings/user_settings.gd
+++ b/addons/godot-xr-tools/user_settings/user_settings.gd
@@ -122,7 +122,7 @@ func save() -> void:
 
 
 ## Get the action associated with a WebXR primary choice
-static func get_webxr_primary_action(primary : WebXRPrimary) -> String:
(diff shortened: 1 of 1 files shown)


squidt:
In the autoload `XRToolsUserSettings` function `get_webxr_primary_action()` is static, which causes `static_called_on_instance` warning when calling from the autoload. As it's only a simple match statement, I chose to unmark it as static. Only called once here: https://github.com/GodotVR/godot-xr-tools/blob/573238b1c224ffb476a8110da320cce04f9ea264/scenes/demo_scene_base.gd#L19

squidt:
Thank you! And if it turns out the static was used, I can change it to an intentional `@warning_ignore` instead.

PR #748, opened 2025-11-12
Fix `return_to_snap_zone` running in editor

squidt:
Simple fix. I was able to reproduce and then fix by adding an `Engine.is_editor_hint()` check in the process function.

image


BastiaanOlij (approved):
Dang you're on a role :)

Yeah good find, that makes perfect sense.

PR #749, opened 2025-11-12
Fix shadowed variables names in interactables

squidt:


BastiaanOlij (approved):
LGTM, this originated in Godot 3 days I think, when we didn't have a position property yet :)

PR #750, opened 2025-11-12
Remove `XRToolsSnapZone` `initial_object` stash sound

squidt:
Stopping the sound after `pick_up_object()` in `_initial_object_check()` allows the stash sound to be suppressed only when that function is used, stopping it from playing during scene loading, but allowing it to play any other time. There is an exception if the AudioStreamPlayer3D has been marked as autoplay by the user. Tested in the `pickable_demo` by making the teacup snap zone play a long and loud audio file with attenuation disabled.

IMPORTANT:
This will still play audio when testing in the `pickable_demo` because of the `XRToolsReturnToSnapZone` picking up the object after a default one-second delay in the editor (which is fixed in #748).

### DIFF
--- a/addons/godot-xr-tools/objects/snap_zone.gd
+++ b/addons/godot-xr-tools/objects/snap_zone.gd
@@ -180,6 +180,13 @@ func _initial_object_check() -> void:
 		# Show highlight when empty and enabled
 		highlight_updated.emit(self, enabled)
 
+	# Stop any audio from initial pickup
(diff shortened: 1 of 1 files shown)


DigitalN8m4r3:
lgtm, btw in my Personal project am using packed scenes as well, what i mean is it might make sense to have a option to load in a packed scene as Initial object instead of one present in the scene, however my proposal here is out of scope

BastiaanOlij (approved):
LGTM

PR #753, opened 2025-11-18
Make sure we get our parent XRController3D node on reparent

### LINKED ISSUE #743: Reparenting functions must update linked controller
Right now we are retrieving the parent controller node in the `_ready` function of most of our scripts.
That means that if nodes are reparented to move a feature from the right hand to the left or visa versa, this has no effect.

Code that retrieves the parent controller will need to be change to implementing `_enter_tree` and `_exit_tree` to ensure we properly deal with reparenting.

BastiaanOlij:
I already fixed this on a few other script but went through the rest.

XR Tools was designed so you can reparent child nodes of the XRController3D nodes to swap left handed/right handed behaviour, however many script initialised on ready and don't react to reparenting.

This PR addresses this.

Do note that the features that work on both hands (like climbing) do not trigger any logic here but I think in that scenario we'll be ok. Needs testing.

Fixes #743

Breaking my own rules here because I was dumb enough not to change branches. Bonus enhancement, you can now bend over tables and configure the playerbody to perform a fade to black when you stick your head through a wall :)

image

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -10,6 +10,7 @@
 - Gracefully handle freed objects with collision exceptions
 - Added support for using grip or palm pose instead of aim pose
 - Various bits of code cleanup
+- Improved player body collisions so players can move over tables
 
 # 4.4.0
 - Minimum Godot version changed to 4.2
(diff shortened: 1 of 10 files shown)


DigitalN8m4r3:
So far lgtm

PR #763, opened 2025-12-13
Better format and static type Area Audio

mubinulhaque:
- Remove space between variable names and colons
- Replace exclamation mark with **not** keyword
- Add static typing to functions **_on_body_entered()** and **_on_body_exited()**

## Disclaimer

### DIFF
--- a/addons/godot-xr-tools/audio/area_audio.gd
+++ b/addons/godot-xr-tools/audio/area_audio.gd
@@ -12,13 +12,13 @@ extends AudioStreamPlayer3D
 
 
 ## XRToolsAreaAudioType to associate with this Area Audio
-@export var area_audio_type : XRToolsAreaAudioType
+@export var area_audio_type: XRToolsAreaAudioType
 
-@onready var area : Area3D = get_parent()
+@onready var area: Area3D = get_parent()
 
 
-# Add support for is_class on XRTools classes
-func is_xr_class(xr_name:  String) -> bool:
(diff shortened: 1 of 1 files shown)


BastiaanOlij:
Sorry for not looking at this sooner. Happy with the changes (and in general with the other PRs as well). Only thing I wouldn't do is remove the lines between the `extends` and the class documentation. I think having the class documentation separated makes it easier to read.

BastiaanOlij:
(though I get the style guide states differently, one of the few things I don't like personally :) )

addons/godot-xr-tools/audio/area_audio.gd
```diff
@@ -12,37 +10,37 @@ extends AudioStreamPlayer3D
...
 
-# Add support for is_class on XRTools classes
-func is_xr_class(xr_name:  String) -> bool:
+## Add support for is_class on XRTools classes
+func is_xr_class(xr_name: String) -> bool:
 	return xr_name == "XRToolsAreaAudio"
 
 
-# Called when the node enters the scene tree for the first time.
+## Called when the node enters the scene tree for the first time.
```
BastiaanOlij: Applies here and in the other PRs too, private functions (those denoted by a _ prefix) don't need public documentation. So a single `#` here is correct.
Only public methods and export variables should have a `##` comment that causes the description to be added to the online help database.

BastiaanOlij (approved):

PR #768, opened 2025-12-13
Better format and static type Controller Hider

mubinulhaque:
- Replace exclamation marks with **not** keyword
- Add variable for parent node to reduce **get_parent()** calls
- Format multiline statements for more readability
- Specify types of variables
- Add spaces before and after equal signs
- Add spaces between if statements
- Rearrange functions appropriately

## Disclaimer


BastiaanOlij (approved):

BastiaanOlij:
Thanks!

PR #776, opened 2026-01-11
Fix Godot 4.x API: params_cull_mode -> cull_mode

thestarfarer:
The `params_cull_mode` property was renamed to `cull_mode` in Godot 4.0.

This only affects the code path when creating material from scratch in `_update_render()` (not when using the bundled `.tscn` which has the material pre-created).


BastiaanOlij (approved):
I'm gonna trust you on that one, we'll see if someone screams :P

PR #787, opened 2026-03-09
Hint for setting default physics layer

VRederikDE:
Just a little hint to the XR Tools tool. Would have saved me some time and hope it helps others.


BastiaanOlij (approved):
Good addition!

Issue #792, opened 2026-04-10
Virtual keyboard does not work in the Pointer demo.


AtomicAntt:
(Using Godot 4.4.1)

Trying to type anything using the virtual keyboard with the ray interactor does not affect the display UI.

The buttons on the virtual keyboard respond and are pressable, but no letters can be written on any lines.

I suspect that the reason for this is because spectator mode creates a sub-viewport, but that prevents inputs from registering onto the display itself when keys were pressed on the virtual keyboard?

PR #793, opened 2026-04-10
Fix virtual keyboard in the pointer demo

AtomicAntt:
Simple change to address https://github.com/GodotVR/godot-xr-tools/issues/792 where we get the Viewport outside of the SubViewport containing the UI and push input there instead of using:

```Input.parse_input_event(input)```

This fixes the virtual keyboard in the demo, as the demo has a SubViewport when implementing the spectator view which seems to block the input event?

One thing to note is that after this change, I still can't edit the lines in the pointer demo with my desktop keyboard, but the virtual keyboard will work.

Works on Godot 4.4.1 and 4.6.2 in the demo, and on another separate Godot XR project (without spectator view).

### DIFF
--- a/addons/godot-xr-tools/objects/keyboard/virtual_keyboard_2d.gd
+++ b/addons/godot-xr-tools/objects/keyboard/virtual_keyboard_2d.gd
@@ -23,6 +23,8 @@ var _alt_down := false
 # Current keyboard mode
 var _mode: int = KeyboardMode.LOWER_CASE
 
+# Viewport node to push input to
+@onready var _viewport: Viewport = get_viewport().get_parent().get_viewport()
 
 # Add support for is_xr_class on XRTools classes
 func is_xr_class(xr_name:  String) -> bool:
@@ -41,10 +43,10 @@ func on_key_pressed(scan_code_text: String, unicode: int, shift: bool):
 	input.pressed = true
 	input.keycode = scan_code
 	input.shift_pressed = shift
-
+	
 	# Dispatch the input event
-	Input.parse_input_event(input)
-
+	_viewport.push_input(input)
+	
 	# Pop any temporary shift key
 	if _shift_down:
 		_shift_down = false
--- a/addons/godot-xr-tools/objects/keyboard/virtual_keyboard_2d.gd
+++ b/addons/godot-xr-tools/objects/keyboard/virtual_keyboard_2d.gd
@@ -23,6 +23,17 @@ var _alt_down := false
 # Current keyboard mode
 var _mode: int = KeyboardMode.LOWER_CASE
 
+# Viewport node to push input to
+var _viewport: Viewport = null
+
+
+func _ready() -> void:
+	# This should be the SubViewport in the virtual keyboard's Viewport2Din3D
+	var vp: Viewport = get_viewport()
+
+	if is_instance_valid(vp.get_parent()):
+		_viewport = vp.get_parent().get_viewport()
+
 
 ## Adds support for [method is_xr_class] on XRTools classes
 func is_xr_class(xr_name: String) -> bool:
@@ -44,6 +55,11 @@ func on_key_pressed(scan_code_text: String, unicode: int, shift: bool) -> void:
 
 	# Dispatch the input event
 	Input.parse_input_event(input)
+
+	# If the viewport of the virtual keyboard's Viewport2Din3D isn't the
+	# root viewport, we should pass the input event to that viewport as well
+	if _viewport != get_tree().root and _viewport != null:
+		_viewport.push_input(input)
 
 	# Pop any temporary shift key
 	if _shift_down:

### CODE REVIEW

**Comment by BastiaanOlij** [comment]
Hmmm, this makes a lot of sense. I do think this should probably become `get_tree().get_root()` so input is always sent to the root viewport and propagated upwards.

Sorry for the late response btw, I've been focusing on XRT2. Need to find some time to properly test this and see if we're not missing anything else but this is definately a fix that should be part of the next release.

**Comment by AtomicAntt** [comment]
Yep, ideally I think it would be best to have the input propagated upwards with this. I believe when I originally made this PR, I also tried pushing input to `get_tree().get_root()`, but it wouldn't work in the XRTools demo specifically (probably because of the spectator view). Unfortunately, my VR headset broke pretty recently, so I can't do any testing relating to this at the current moment :(

Also completely understand the need to focus on XRT2, I am pretty excited for its release actually.

**Comment by mubinulhaque** [comment]
I tested this code and it does fix the issue.

I also agree with @AtomicAntt's approach here. Sending it to the root Window won't fix the issue, as we want the input to parsed by HMDViewport. Unfortunately, while calling `get_parent()` is messy, it is probably the best way of handling a spectator view.

**Comment by AtomicAntt** [comment]
Thanks @mubinulhaque for testing the code out for me. Actually I recently managed to get a replacement VR headset from Meta so I can get back into making any changes to this if needed.

Regarding this fix, actually it might not be a bad idea to contain input in the HMDViewport instead of propagating upwards, because it could be possible for input fields to be in the spectator view that the user doesn't want interacted with via virtual keyboard? I'm not too familiar with how spectator view is implemented in the demo so do correct me if I'm wrong.

**Comment by mubinulhaque** [comment]
Input is already parsed by the HMDViewport with your code, unless that's not what you meant by 'contain'? Regardless, while this solution works, perhaps a better one would be to send it to the root viewport and then propogate it downwards somehow via the spectator camera script? Is that possible in Godot?

@AtomicAntt the spectator camera and script are called `demo_spectator.tscn` and `demo_spectator.gd` respectively, if you want to familiarise yourself with it.

**Comment by AtomicAntt** [comment]
@mubinulhaque What I meant was that this solution will push input into the HMDViewport if one exists, and that actually propagating it upward towards the root viewport may actually be undesired because of the side effects.

For example, if someone wanted to make a kind of local co-op game, where someone on the desktop could write words on the spectator view with a control node like a `LineEdit`, using the virtual keyboard to push input there might be undesirable. (With a quick test, I can confirm that this would occur)

I was thinking if it would be desirable to prevent similar things like that from happening, or if it would be preferrable to propagate the input upwards from the virtual keyboard. The solution here currently would keep the input "contained" inside the HMDViewport or just the root viewport if no spectator view exists.

As for propagating it upwards, I can probably make it so instead it can keep looking for viewport ancestors, then pushing input, using `get_parent().get_viewport()`, and stopping once I hit the root viewport, or keep this method and also push input to `get_tree().get_root().get_viewport()` if it isn't already the root viewport.

**Comment by mubinulhaque** [comment]
Giving the user the option to either prefer the HMD viewport or the root viewport for input handling sounds like a good idea to me.

**Comment by AtomicAntt** [comment]
> Giving the user the option to either prefer the HMD viewport or the root viewport for input handling sounds like a good idea to me.

@mubinulhaque Hmm, actually I do like that idea because users can choose how they want input to be handled.

One thing that makes this a little bit tricky though is that `virtual_keyboard_2d.gd` is a PackedScene, so `@export` variables will not be as accessible when users use the virtual keyboard scene. (See addons/godot-xr-tools/objects/virtual_keyboard.tscn)

And if we do make it an option, I wonder which option should be the default one.

**Comment by mubinulhaque** [comment]
I have two proposals, but they're both messy. 

1. Make a new child class of `Viewport2Din3D` called `virtual_keyboard.gd` and add the option to choose the viewport to parse input there
2. Add the option to choose a viewport to parse input in `Viewport2Din3D`

**Comment by AtomicAntt** [comment]
After thinking about it, I think it's best to keep `Input.parse_input_event(input)` so that developers trying to process input in `_process()` will keep working when they use this virtual keyboard.

After testing, having this in an autoload wouldn't work anymore for example (Even without spectator mode, which was the issue here):
```gdscript
func _process(delta) -> void:
    if Input.is_action_just_pressed("ui_accept"):
        print("Accept")
```

For this reason, I have updated the code to keep using `Input.parse_input_event(input)`, and then only pushing input to the outer viewport when necessary, which will ensure compatibility.

Tested with spectator mode and without spectator mode on Godot 4.4.1 and in Godot 4.7.stable.

addons/godot-xr-tools/objects/keyboard/virtual_keyboard_2d.gd
```diff
@@ -23,6 +23,8 @@ var _alt_down := false
 # Current keyboard mode
 var _mode: int = KeyboardMode.LOWER_CASE
 
+# Viewport node to push input to
+@onready var _viewport: Viewport = get_viewport().get_parent().get_viewport()
```
BastiaanOlij: This should move to a normal `ready` function so you can write out the code and check for valid nodes.
There is a risk here of the code erroring out if it's accidentally run on the main viewport (very unlikely but not impossible).

AtomicAntt: Oh yeah, guess that's true. I've updated it to check if the viewport has a valid parent first (which the main viewport wouldn't have) before initializing that viewport to push input into. And some clarifying comments

BastiaanOlij (approved):
LGTM

PR #799, opened 2026-06-06
Better format Pointer Event

mubinulhaque:
- Remove space between variable names and colons
- Add class doc comment
- Clarify two pointer events' comments
- Reorder functions

## Disclaimer


BastiaanOlij (approved):
LGTM

PR #800, opened 2026-06-06
Better format and static type Fall Damage

mubinulhaque:
- Add return type to function **physics_movement()**
- Replace `emit_signal` with `signal.emit()`
- Add type to signal parameter
- Reorder functions

## Disclaimer


BastiaanOlij (approved):
LGTM

PR #802, opened 2026-06-11
Better format and static type Gaze Pointer

mubinulhaque:
- Add type to signal `pointing_event` parameter
- Reorder functions in order of public before private, then alphabetically
- Add suffixes to five exported properties

## Disclaimer

### DIFF
--- a/addons/godot-xr-tools/functions/function_gaze_pointer.gd
+++ b/addons/godot-xr-tools/functions/function_gaze_pointer.gd
@@ -3,29 +3,26 @@
 class_name XRToolsFunctionGazePointer
 extends Node3D
 
-
 ## XR Tools Function Gaze Pointer Script
 ##
 ## This script implements a pointer function for a players camera. Pointer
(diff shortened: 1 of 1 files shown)


mubinulhaque:
As a quick aside, while this is out of scope for this PR, I took a quick diff and noticed that a lot of this code is the same as the code from XRToolsFunctionPointer. Would it be possible for us to simply make a new class that contains all the repeated code and add these to both classes?

BastiaanOlij:
> As a quick aside, while this is out of scope for this PR, I took a quick diff and noticed that a lot of this code is the same as the code from XRToolsFunctionPointer. Would it be possible for us to simply make a new class that contains all the repeated code and add these to both classes?

It would probably be worth it to see if we could construct a common base class, they indeed have a lot in common.

BastiaanOlij (approved):
LGTM, needs some testing just in case moving methods around caused some issues.

BastiaanOlij:

PR #803, opened 2026-06-11
Fix _property_get_revert() throwing error

mubinulhaque:
I added an `else` clause for every time _property_get_revert() is called. This should fix Godot 4.7rc1 throwing an error whenever XRToolsFunctionTeleport or Viewport2Din3D is used. The demo should work as normal (although I did encounter one weird bug where teleporting onto a slope causes you to slide **up** a slope, rather than down it, but this behaviour occurs in `master` anyway).


BastiaanOlij (approved):
LGTM, thanks for taking the time to make these changes!

PR #804, opened 2026-06-13
Better format and static type Pickup Function

mubinulhaque:
- Add type to several procedure parameters
- Add type to signal `has_picked_up` parameter
- Reorder functions according to the style guide above
- Add two export groups

## Testing
A lot of functions have been moved around, so I tested both the Pickables demo and the Audio demo, both of which run the same as `master`.

## Note
Originally, I cast several variables and parameters as `XRToolsPickable`, but this spammed errors, since `XRToolsSnapZone`s can also be picked up. Hence, I cast these variables as `Node3D` instead, losing a bit of type safety. If we want this to change, we'll need to modify `XRToolsSnapZone`, which is out of scope for this PR.

Also, if we ever push the minimum version of Godot for XR Tools to 4.6+, we should use `PROPERTY_HINT_GROUP_ENABLE` for the export group `Ranged Grab`.

## Disclaimer

### DIFF
--- a/addons/godot-xr-tools/functions/function_pickup.gd
+++ b/addons/godot-xr-tools/functions/function_pickup.gd
@@ -3,7 +3,6 @@
 class_name XRToolsFunctionPickup
 extends XRToolsHandPalmOffset
 
-
 ## XR Tools Function Pickup Script
 ##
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):

PR #805, opened 2026-06-13
Better format and static type Return To Snap Zone

mubinulhaque:
- Reorder functions according to the style guide above
- Add types to method parameters
- Add suffix to exported property

## Disclaimer


BastiaanOlij (approved):

PR #807, opened 2026-06-13
Better format and static type Pointer

mubinulhaque:
- Add type to signal `pointing_event` parameter
- Add suffixes to three exported properties

## Testing
The Pointer demo was used for testing. Everything works as intended, although I used an await in the `_enter_tree()` method, otherwise this throws an error. AFAIK, this should not change anything. However, I should note that the keyboard in the Pointer demo does not do anything, even in `master`, due to issue #792.

## Disclaimer


BastiaanOlij (approved):

Issue #584, opened 2023-12-10
Interactables need 3D Gizmos to assist in alignment


Malcolmnixon:
The Interactables (Hinge, Slider, Joystick) need [3D Gizmos](https://docs.godotengine.org/en/stable/classes/class_editornode3dgizmoplugin.html#class-editornode3dgizmoplugin) to assist in alignment and configuration; however the design of the interactables currently hinders this.

The current interactable nodes move within the confines/transform of their parent "Origin" Node3D; and so visualizing the extents of motion want to be with regards to that "Origin" Node3D; however the 3D Gizmos mechanism attaches visualizations and aids to the target node.

As such a secondary set of scripts HingeOrigin, SliderOrigin, JoystickOrigin would be needed to define the extents, and the 3D Gizmos would be attached to these origin nodes.

PR #808, opened 2026-06-15
Added gizmos to help align interactable joints

mubinulhaque:
# Note
Supersedes #585. Co-authored-by: Malcolm Nixon <Malcolm.nixon@gmail.com>

# Original PR
This PR resolves request #584 by adding gizmos to align interactable hinges, sliders, and joysticks. This does cause minor breakage in that:
- New origin nodes have been introduced where the ranges are visualized
- Testing showed Y-Axis inversion on the slider and joystick

# Demonstration Video
https://github.com/GodotVR/godot-xr-tools/assets/1863707/0bf443f5-5b3b-48b5-b953-c7ca4d866196

# Slider Gizmo
image

# Hinge Gizmo
image

# Joystick Gizmo
image

# Disclaimer
No generative AI was used **by me** to enhance or create the code given here.

### DIFF
--- a/addons/godot-xr-tools/editor/gizmos/hinge_origin.gd
+++ b/addons/godot-xr-tools/editor/gizmos/hinge_origin.gd
@@ -1,0 +1,185 @@
+extends EditorNode3DGizmoPlugin
+
+## Editor Gizmo for [XRToolsInteractableHingeOrigin]
+##
+## This editor gizmo helps align interactable hinge joints.
+
+var undo_redo: EditorUndoRedoManager
+
+
+func _init() -> void:
+	create_material("axis", Color(1, 1, 0))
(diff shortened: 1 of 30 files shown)


BastiaanOlij:
I'd have to look into this a little closer, is it not possible to put this logic on the interactable nodes themselves?
I imagine this will be trickier because those nodes need to be able to move in relation to the origin node?

Possibly something we'll have to accept for XR Tools but redesign once interactable are added to XR development kit.

mubinulhaque:
A lot of this logic requires the use of the class `EditorNode3DGizmoPlugin`, so it's just easier to move a lot of the logic to new classes, rather than to try shoehorning it into existing classes and breaking things.

PR #773, opened 2025-12-15
Better format and static type Desktop Flight

mubinulhaque:
In accordance with [style guide given here](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html):
- Add doc comments to variables and functions where possible
- Replace exclamation marks with **not** keyword
- Format multiline statements for more readability
- Add spaces before and after equal signs
- Remove unnecessary _ready() function
- Specify return type of function **physics_movement()**
- Replace **emit_signal("signal")** with **signal.emit()**

## Disclaimer
No generative AI was used to enhance or create the code given here.

PR #772, opened 2025-12-15
Better format and static type Desktop Direct Move

mubinulhaque:
In accordance with [style guide given here](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html):
- Add doc comments to variables and functions where possible
- Replace exclamation marks with **not** keyword
- Format multiline statements for more readability
- Specify return type of function **physics_movement()**

## Disclaimer
No generative AI was used to enhance or create the code given here.

PR #813, opened 2026-06-19
Change physics_movement() return type to bool

mubinulhaque:
Back when I made the cleanup PRs for the Desktop Movement Providers (#771, #772, #773, #779, #780, #782), I incorrectly assumed that the return type for `physics_movement()` was supposed to be `void`, since they would always return a null value. Looking at the `XRToolsPlayerBody` class properly and talking with Bastiaan Olij on Discord, it seems that it's actually supposed to be `bool`, so I made the correct changes before 4.6.0 is released.
# Note
This is a **breaking** change! I had to add `return false` statements at the end of some of these methods, so some stuff may break!


BastiaanOlij (approved):
Important for compatibility with 4.7, looks good, will fully test once all is merged.

PR #817, opened 2026-06-20
Better format and static type Footstep

mubinulhaque:
- Add return type to two functions
- Rename a signal parameter and defines its type as `String`
- Change the parameter `name` in `_play_sound()` to `surface` for better clarity
- Add suffixes to two exported properties
# Note
A portal to the **Footstep Movement** demo is not available in the main menu scene. A PR (#818) to add this has already been opened.
# Testing
The **Footstep Movement** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #818, opened 2026-06-20
Clean up and add footstep demo to main menu

mubinulhaque:
This PR does the following:
- Adds a portal to the main menu that leads to the footstep movement demo
- Swapped two of the small tiles around to be closer to their bigger counterparts
- Added text atop the big tiles that describes what the surfaces are supposed to be
- Changed the mud tile to be more brown
- Moved the instructions closer to the centre to fix the flickering issue
# Testing
The **Footstep Movement** demo had no issues not present in `master`. However, the texture used for the portal was used as is, and does not match the other portal textures. Should we change it to be more in line?
# Disclaimer


BastiaanOlij (approved):

PR #819, opened 2026-06-20
Better format and static type Glide

mubinulhaque:
- Add BBCode to the class documentation
- Add return type to the function `physics_movement()`
- Add two export groups
- Add suffixes to exported properties where appropriate
# Note
If we ever push the minimum version for XR Tools to 4.6+, we could use `PROPERTY_HINT_GROUP_ENABLE` for the export groups.
# Testing
The **Climbing & Gliding** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #821, opened 2026-06-21
Better format and static type Jog

mubinulhaque:
- Add suffixes to two exported variables
# Note
This is one of the few PRs that did not reorder functions around.
# Testing
The **Sprinting** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #837, opened 2026-06-27
Better format and static type Move To

mubinulhaque:
- Replace `connect(signal, Callable)` with `signal.connect(Callable)`
# Note
I couldn't find any scene that uses this Node, and this script looks incomplete, considering that `duration` is never set anywhere. As such, while I have updated it, I cannot test it anywhere.
# Disclaimer


BastiaanOlij (approved):

PR #840, opened 2026-06-28
Static type VR Shader Cache

mubinulhaque:
- Add types to variables
- Add comments to a variable and a signal
# Testing
The **Main Menu** scene had no issues not present in `master`. Note that more testing should be done on a fresh machine that has never loaded the Shader Cache before.
# Disclaimer


BastiaanOlij (approved):

PR #841, opened 2026-06-28
Better format and static type XR Helpers

mubinulhaque:
- Add BBCode to class doc
- Add type to `for` loop variable
# Testing
The **Main Menu** scene and the **Pointer** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #842, opened 2026-06-28
Better format and static type Force Body

mubinulhaque:
- Add BBCode to class doc
# Testing
The **Pickables** demo had no issues not present in `master`. Do note that currently `master` has an issue where, after placing the scope on the sniper rifle and placing the rifle in your Snap Zone, the scope will be misaligned as the player moves around.
# Disclaimer


BastiaanOlij (approved):

PR #844, opened 2026-06-28
Better format Grabber

mubinulhaque:
- Remove doc comment for private method
# Testing
I couldn't find any place where the class `Grabber` was used. Nevertheless, the **Pickables** demo and the **Interactables** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #854, opened 2026-06-30
Better format and static type Virtual Keyboard

mubinulhaque:
- Clarify comments of variables
# Note
This will cause merge conflicts with #793.
# Testing
The **Pointer** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (commented):

BastiaanOlij (approved):

PR #858, opened 2026-06-30
Better format and static type Interactable Area

mubinulhaque:
- Add class doc
- Add type to signal `pointer_event` parameter
- Clarify comments of methods and signals
# Testing
This is not used anywhere in the Godot XR Tools demo. In the future, this should be added to the **Pointer** demo.
# Disclaimer


BastiaanOlij (approved):

PR #860, opened 2026-07-01
Better format and static type Pickable

mubinulhaque:
- Add return type to nine functions
- Add type to `Dictionary`
# Note
We are using a `Dictionary` for highlight requests, when we should really be using an `Array`. This is out of scope for this PR, though.
# Testing
The **Pickables** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #861, opened 2026-07-01
Better format and static type Snap Zone

mubinulhaque:
- Add return type to three functions
- Add type to `Array`
# Note
I wanted to set the type of the `Array` to `XRToolsPickable`, but considering several methods check for signals that this class should have, I decided to err on the side of caution and set it to `Node3D`. If this is wrong, please let me know.

I also set the default radius in the `snap_zone.tscn` scene to be 0.3, simply because that's the default in `snap_zone.gd` as well.
# Testing
The **Pickables** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #863, opened 2026-07-01
Better format and static type Viewport 2D in 3D

mubinulhaque:
- Add return type to six functions
# Note
This commit also includes changes to the default values in the scene, that are brought over from the default values in the script (which were left unchanged).

For the Physics export group, we should consider using `PROPERTY_HINT_GROUP_ENABLE` if we ever change the default Godot minimum version to 4.6+.

This is a big PR, so a lot of testing will be needed.
# Testing
The **Pointer** demo had no issues not present in `master`. Do note that `master` currently has a bug where the player cannot 'poke' UI elements while gazing at them, probably due to interference from the Gaze Pointer.
# Disclaimer


BastiaanOlij (approved):

PR #867, opened 2026-07-02
Better format and static type Ground Physics Settings

mubinulhaque:
- Make physics overrides read-only when their respective flags are unchecked
# Testing
The **Sprinting** demo had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #868, opened 2026-07-03
Better format and static type Fade Collision

mubinulhaque:
- Add suffixes to two export variables
# Testing
The **Basic Movement** demo had no issues not present in `master`. Do note that `master` currently has a bug where it is possible to peek through the wall and see the outside world clearly.
# Disclaimer


BastiaanOlij (approved):

PR #875, opened 2026-07-05
Better format and static type Loading Screen

mubinulhaque:
- Add return type to seven methods
- Add BBCode to class doc
# Note
There are two versions of `loading_screen_shader.tres`, only one of which is used. Shouldn't one be deleted?
# Testing
The **Loading Screen** scene had no issues not present in `master`.
# Disclaimer


BastiaanOlij (approved):

PR #877, opened 2026-07-05
Better format and static type Staging

mubinulhaque:
- Add return type to ten methods
- Add BBCode to class doc
- Replace two repeated node paths with variables
- Replace `emit_signal(signal, Callable)` with `signal.emit(Callable)`
- Replace `connect(signal, Callable)` with `signal.connect(Callable)`
- Replace `disconnect(signal, Callable)` with `signal.disconnect(Callable)` 
# Note
I've left the variable `user_data` as Variant, since that seems to be custom data that can be overridden by the user.
# Testing
# Disclaimer

### DIFF
--- a/addons/godot-xr-tools/staging/staging.gd
+++ b/addons/godot-xr-tools/staging/staging.gd
@@ -2,113 +2,112 @@
 class_name XRToolsStaging
 extends Node3D
 
-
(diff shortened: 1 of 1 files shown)


BastiaanOlij (approved):

PR #881, opened 2026-07-16
Use `GridContainer`s for User Settings UI instead of `HBoxContainer`s

mubinulhaque:
To decrease the amount of nodes, and to make creating new settings easier, I have replaced the several `HBoxContainers` with a single `GridContainer` per section. I have also changed the `Label`s for the Apply buttons to "Save" instead, which is more in line with what these buttons do.
# Testing
# Disclaimer


BastiaanOlij:
I'll give this a proper try once rebased, sounds like a nice fix based on the description.

mubinulhaque:
Rebased!

PR #882, opened 2026-07-16
Better format and static type Start XR

mubinulhaque:
- Add return type to two methods
- Add BBCode to class doc
- Replace `emit_signal(signal, Callable)` with `signal.emit(Callable)`
- Replace deprecated methods where possible
   - Line 280 still contains usage of a deprecated method, because no other method is able to replace it
# Testing
# Disclaimer


BastiaanOlij (approved):

PR #883, opened 2026-07-21
Better format and static type Plugin script

mubinulhaque:
- Add return type to two methods
- Properly format multiline statements
- Remove checking if `ProjectSettings` has the method `set_as_basic()`, since this is always included in 4.4+
# Note
Line 108 is checking if the Project Settings contains the setting, and if **not**, sets the value for that setting. Is this intended behaviour? Should it not set the value only **if** the setting exists?
# Testing
I tried adding this plugin to a fresh new project, and I didn't get any errors.
# Disclaimer


BastiaanOlij (approved):

PR #885, opened 2026-07-21
Add grip rotation override

mubinulhaque:
Adds two new Project Settings
- Toggle for overriding grip rotation
- What the override for the grip rotation should be (default is -45 degrees)
# Testing
Adding the plugin to a fresh new project contained no issues not present in `master`.
# Disclaimer

### DIFF
--- a/addons/godot-xr-tools/plugin.gd
+++ b/addons/godot-xr-tools/plugin.gd
@@ -137,6 +137,24 @@ func _enter_tree():
 			"1.0,2.5,0.05",
 			1.85)
 
+	# Add an toggle for overriding grip rotation
+	_define_project_setting(
+			"godot_xr_tools/input/grip_rotation_override",
+			TYPE_BOOL,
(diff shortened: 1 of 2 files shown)


BastiaanOlij:
What is the use case we're going for? The original idea was to complete the list of rotation offsets so they are correct for each platform (and hopefully long term switch to using the more reliable palm pose).

I want to make sure we don't have users think they must set this and lock themselves into a single dev platform, but I can see the use case if users want to finetune the rotation in a user menu.

mubinulhaque:
Allowing users to fine tune this in a setting is absolutely a use case, since we can disable overriding the grip rotation by default. Furthermore, many users may be transferring their projects over from old versions of XR Tools where the grip pose was preferred, and they may not have the time to go back and change this to the more reliable palm pose without breaking anything during testing.

If you'd like, I can add a comment in the Project Setting telling developers that this is intended for finetuning the grip rotation in a user menu.

PR #891, opened 2026-09-17
Fix distant pickup

renaiku:
The PR #804 introduced a bug in the function pickup  `_on_ranged_entered()` by rejecting every Node without a `can_ranged_grab` function.

I think most of the scenes, like the basketball, relies on a var `can_ranged_grab`, and they are rejected.

Can you confirm @mubinulhaque ?
(btw thank you for all these typings everywhere 😍)


mubinulhaque:
As far as I can see, the code is the same, and the part that says `or not target.can_ranged_grab` should be checking for the variable.

Nevertheless, I checked out your branch, and couldn't find any difference. Which demo are you testing?

renaiku:
I've updated my addon version in my project and this change broke the distant grab.

Indeed when reading again my commit, it should be better, I will rework it, because it just reversed the problem but with the variable instead.

The new code with typings introduced that if the pickable doesn't have the function, it will return (even if it has the variable)

```
if not target.has_method('can_ranged_grab') or not target.can_ranged_grab:
    return
```

mubinulhaque:
I see what it is now. Because of the line `if not target.has_method('can_ranged_grab') or not target.can_ranged_grab:`, it immediately returns once the method is not found.

This should just be changed to `if not "can_ranged_grab" in target or target.can_ranged_grab:`. We should **not** cast `target` to `XRToolsPickable`, because then that would exclude `XRToolsSnapZone`s.

renaiku:
Oh nice catch for the snap zone. Should I update the PR with your code proposal or I met you handle this and I close this PR ?

mubinulhaque:
Updating the PR you have here is fine.

mubinulhaque:
There is a problem with the new changes: we are now rejecting ranged grabs for objects that explicitly enable them. Instead of:
`if not "can_ranged_grab" in target or target.can_ranged_grab:`
we should put:
`if not "can_ranged_grab" in target or not target.can_ranged_grab:`

renaiku:
damned I copy pasted what you wrote without reading 🥲

mubinulhaque:
Can you [squash your commits](https://contributing.godotengine.org/en/latest/pull_requests/creating_pull_requests.html#modifying-a-pull-request)?

PR #892, opened 2026-09-17
Fix `highlight_updated` causing errors

mubinulhaque:
It seems that both `XRToolsSnapZone` and `XRToolsPickable` have a signal called `highlight_updated`. I had mistakenly assumed that only the latter had the signal, which meant that the highlight classes were connecting to these signals and trying to cast the Snap Zones as Pickables, as shown below:
> ERROR: core/object/object.cpp:1249 - Error calling from signal 'highlight_updated' to callable: 'MeshInstance3D(highlight_ring.gd)::_on_highlight_updated': Cannot convert argument 1 from Object to Object.

I have now corrected the typing in the signals to instead have the pickable as `Node3D`, rather than `XRToolsPickable`.
# Testing
The `Pickables` demo had no issues not present in `master`. Reloading the project with this scene now no longer causes errors.
# Disclaimer


mubinulhaque:
A quick code search shows that the first parameter `pickable` of the `highlight_updated` signal is never used. Would it be worth removing, or is it intended for users?

